PR #1, opened 2017-01-23
Desktop styled bar

m4dz:


m4dz:
@GoOz, I've got an issue when including SVG icons from the sprite. I can't figure why. If you can take a look at it, it'll be a great appreciated help ❤️

GoOz:
@m4dz So far I can't tell why those SVGs aren't displayed. Did you get any luck since?

m4dz:
@GoOz unfortunaltely: nope. This is my mission for the day 💣

GoOz:
@m4dz Feel free to poke me if you need a new eye on this 😄

src/styles/bar.css
```diff
@@ -5,7 +5,22 @@ body {
...
 
+
 [role=application] {
   height: initial;
   flex: 1 100vh;
 }
+
+
+[role=banner] {
+  --silver: #d6d8da;
```
CPatchane: CSS variables seems to be not very supported ([only in very recent browser](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_variables)), are they necessary here? /poke GoOz

GoOz: @CPatchane Yes, I pointed that out too on the mobile branch as well.

m4dz: We normally enable a fallback using autoprefixer, but it seems to not work here, which is a little bit annoying. I'll investigate, thanks.

GoOz: @m4dz Are you sure autoprefixer supports custom properties? I tried using CSSNext syntax on a project once and I had to use the postcss plugin **cssnext** to make it work. 
CSSNext plugin brings autoprefixer with it so i'm not sure which does what but here's what I know. 👅

m4dz: @GoOz you're right, my mistake, we'll pass through a postcss plugin to handle them 😉

m4dz:
As I'm working on the optimizations right now on another branch (css variables, build chain, mq packing, etc), can we consider this one as valid and merge it? So I can reassemble my branches asap. Thanks!

GoOz:
👍

PR #2, opened 2017-01-23
Mobile styles

m4dz:
**⚠️ THIS ONE AWAIT FOR #1, DO NOT MERGE BEFORE IT**

Add mobile / rwd styles and markups


src/components/Drawer.svelte
```diff
@@ -0,0 +1,130 @@
...
+
+  .coz-drawer-wrapper::before {
+    content: '';
+    display: block;
+    position: absolute;
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+    background-color: var(--charcoal-grey);
```
GoOz: I noticed that you used custom properties (nice 😉) but aren't we supposed to support Edge browser? Because, afaik it doesn't support custom properties or not fully anyway

src/components/Bar.svelte
```diff
@@ -0,0 +1,97 @@
...
+  .coz-bar-title img {
+    margin-right: .45em;
+  }
+
+  .coz-bar-title strong {
+    margin-left: .25em;
+    font-weight: bold;
+  }
+
+  @media (max-width: 30em) {
```
m4dz: I'me used to declare media-queries that directly follow the component they modify, as it seems to be more simple to debug / maintain. Postcss will they pack all MQ at the end when the final CSS is built. Does this sounds good, or do you prefer we adopt another convention?

src/components/Drawer.svelte
m4dz: Yep, I expect to handle them with postcss, I'm working on it right now 👍

m4dz:
@GoOz about `<hr />` tag, there's here about separation of parts of content, but I understand your pov. What do you recommend to use insted?

src/components/Bar.svelte
GoOz: aaaah ok I get it.
No, it's ok. I misread, I thought the two media queries were adjacent.
It's fine like this. 😄

GoOz:
@m4dz Nothing, keep it that way.
I read again HTML5Doctor's interpretation of the spec and it's still fuzzy (well you know how are specs…) but it's quite good enough semantically. If it's debatable, it's good enough 😛

m4dz:
> If it's debatable, it's good enough

❤️ this one! I rebase and it'll be good to merge 👍

m4dz:
@GoOz ok, ready to merge for me 😄

GoOz:
👍

PR #6, opened 2017-02-02
Apps nav

m4dz:
This PR enables the app nav in CozyBar.

Features:

- [x] Get apps list from server stack
- [x] Deal with multiple dropdowns
- [x] Display mobile view

Disabled on this PR:

- Handle issues from networks resources
- Handle issues from no-apps in stacks~~ (as viewed with @kelukelu, we consider we always have at least one app, so this use-case can't occur)

Currently, this codebase doesn't handle two points currenlty unavailable in stack:
- path to the app icon
- url to open (in another tab?) to launch app


src/components/Bar.svelte
```diff
@@ -9,25 +9,54 @@
...
 
   import Navigation from './Navigation'
   import Drawer from './Drawer'
 
   import MENU_CONFIG from '../config/menu'
 
+  async function updateAppsItems () {
+    try {
+      let apps = await stack.get.apps()
+      apps = [await apps.map(app => {
```
CPatchane: For my culture, is it necessary to use await here for a simple map usage here?

m4dz:
Merged in #9, close it

PR #7, opened 2017-02-06
Enable multiple savors in build

m4dz:
This PR split webpack config to enable multiple savors when building the the cozy-bar lib:

- `cozy-bar.{js,css}` : the multiple assets version, to be injected by the stack (see https://github.com/cozy/cozy-stack/pull/260)
- `cozy-bar.standalone.js` : the standalone version ready to use that auto-inject the styles in the `head` section

both are also exported in a minified version.


aenario:
@jinroh wont be available for some times, @gregorylegarec feel free to merge if you're ok with it.

PR #10, opened 2017-02-08
Add network issues handler

m4dz:


m4dz:
@CPatchane I've made some little improvement to the Storage component prior to my work on the apps dropdown to mutualize spinner styles and behaviors. Let's discuss about it, I'm aware on all your feedback :smile:

m4dz:
@CPatchane I've made some little improvement to the Storage component prior to my work on the apps dropdown to mutualize spinner styles and behaviors. Let's discuss about it, I'm aware on all your feedback :smile:

src/components/NavigationItem.svelte
```diff
@@ -25,6 +22,13 @@
 
   export default {
     computed: {
+      isBusy: item => {
+        if (!item.component) { return false }
+
+        if (item.component === 'storage') {
+          return !item.currentDiskUsage
```
CPatchane: If the currentDiskUsage is 0 (which can be possible), the spinner will be continuously here even if the stack request succeeds. I think we should do: `return item.currentDiskUsage !== false` to check boolean type as well.

PR #13, opened 2017-02-17
Apps icons and minor improvements

m4dz:
This PR add proper loading for apps icons in cozy-bar, w/ a lazy loader that allow us to fallback to a placeholder icon if the given URL returns an error (like a 404).

It also brings some improvement to lightweight final build.

### DIFF
--- a/src/components/NavigationSection.svelte
+++ b/src/components/NavigationSection.svelte
@@ -1,9 +1,9 @@
 <li class='coz-nav-section' on:click='dispatch(event)'>
-  <a on:click='toggle()' aria-controls='{{`coz-nav-pop--${slug}`}}' aria-busy='{{busy}}' data-icon='{{icon}}'>
+  <a on:click='toggle()' aria-controls='{{`coz-nav-pop--${hash}`}}' aria-busy='{{busy}}' data-icon='{{icon}}'>
     {{t(label)}}
   </a>
   {{#if items}}
-  <div class='{{`coz-nav-pop coz-nav-pop--${slug}`}}' id='{{`coz-nav-pop--${slug}`}}' aria-hidden={{closed}}>
+  <div class='{{`coz-nav-pop coz-nav-pop--${hash}`}}' id='{{`coz-nav-pop--${hash}`}}' aria-hidden={{closed}}>
     {{#if items.error}}
     <p class='coz-nav--error coz-nav-group'>
       {{t(`error_${items.error.name}`)}}
@@ -18,7 +18,6 @@
 </li>
 
 <script>
-  import slug from 'slug'
   import { t } from '../lib/i18n'
 
   import NavigationGroup from './NavigationGroup'
@@ -77,7 +76,7 @@
       }
     },
     computed: {
-      slug: label => slug(label)
+      hash: label => Math.abs([].reduce.call(label, (hash, char) => char.charCodeAt(0) + (hash << 5) + (hash << 16) - hash, 0))
     },
 
     onrender () {
--- a/src/index.js
+++ b/src/index.js
@@ -18,6 +18,12 @@ const injectDOM = function CozyBarInjectDOM (data) {
 
   require('./styles/index.css')
 
+  try {
+    document.matches('[role=foo]')
+  } catch (e) {
+    return console.warn("Cozy-bar is looking for a `[role=application]` tag that contains your application and can't find it :'(… The BAR is now disabled")
+  }
+
   const barNode = createElement()
   const appNode = document.querySelector('[role=application]')
   document.body.insertBefore(barNode, appNode)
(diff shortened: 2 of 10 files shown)


src/styles/navigation_section.css
```diff
@@ -63,18 +63,19 @@
   display: none;
 }
 
-[role=banner] .coz-nav-pop--apps .coz-nav-group {
+/* coz-nav-pop--apps label is hashed coz-nav-pop--1612761714 */
+[role=banner] .coz-nav-pop--1612761714 .coz-nav-group {
```
CPatchane: The hash will always be the same here? What's the interest? Just for my culture.

src/components/NavigationSection.svelte
```diff
@@ -77,7 +76,7 @@
       }
     },
     computed: {
-      slug: label => slug(label)
+      hash: label => Math.abs([].reduce.call(label, (hash, char) => char.charCodeAt(0) + (hash << 5) + (hash << 16) - hash, 0))
```
CPatchane: Can you explain me that or if you have a source just for my understanding here. Thanks 👍

src/styles/navigation_section.css
m4dz: Yep, hash will always be the same, it's a checksum from the label. I don't want to use the label itself as it can be inconsistent (spaces, diacritics, etc). I first used the `slug` module bu it comes with a ~1.5Mb dictionnary, bundled into the lib, so it's clearly a no-go :tongue:

I fallback to a simple checksum (cf https://github.com/cozy/cozy-bar/pull/13/files#diff-8f646533758ab8608cfd8140b4c1297cR79) based on the label. It's lightweight and as consistent as we need (very similar to CSS modules, in fact)

src/components/NavigationSection.svelte
m4dz: It's a simple an fast checksum algorithm initially created for sdbm (an OSS reimplementation of NDBM). I just wanted a simple way to generate a checksum without the need to embed an _md5_ or _sha_ module, and without rely on webcrypto which is clearly too complicated (promises and arraybuffers) for our need here: we just need a checksum based on the label after all :tongue:

This one is fast, any other can be as nice as this one, we can switch when / if we need later :smile:

src/index.js
```diff
@@ -27,6 +28,10 @@ const injectDOM = function CozyBarInjectDOM ({lang, appName, iconPath}) {
   })
 }
 
+const getDefaultStackURL = function GetDefaultCozyURL () {
+  return document.querySelector('[role=application]').dataset.cozyDomain
```
gregorylegarec: Imagine that some contributors try the cozy-bar for the first time but did not start their application with our template and do not have a `[role=application]` element. Shouldn't we provide et better experience by throwing a clear error like "Your application need an element having with a role=application attribute" ? Here it will just crash.

src/index.js
```diff
@@ -22,12 +22,21 @@ const injectDOM = function CozyBarInjectDOM ({lang, appName, iconPath}) {
   const appNode = document.querySelector('[role=application]')
```
gregorylegarec: Same as previous comment : what if '[role=application]' element does not exist ?

src/styles/navigation_section.css
gregorylegarec: For my culture, what's the point of having a hash here ?

src/index.js
m4dz: As the BAR is dedicated to be used in cozy's apps, I assume we can rely on the `[role=application]` tag, as this one is in our base template.

Eventually, we can add an option in the BAR's `init` method, tht allow any developer to specify a CSS selector for its app's root tag, and fallback to `[role=application]` by default?

m4dz:
About testing: we haven't any simple solution to test as I hadn't fixed tests on https://github.com/cozy/cozy-files-v3/pull/110 yet. After this, it'll be more simple to test libs on local versions.

About hash: we rely on label, that can be uncompliant with CSS classnames (spaces, diacritics, etc). I previoulsy use a slugification method, but it brings a ~1.5Mb dictionnary bundled into the lib, which is clearly a no-go. So I fallback to a checksum, which is consistent, lightweight, and faster :+1:

gregorylegarec:
For testing, I was thinking about a cozy-bar standalone version :) I do not really know how you work on this cozy-bar.

My question about the hash/checksum/key was more like: what is his role ?

src/index.js
gregorylegarec: Ok my primary question about `[role=application]` was into this outdated comment @m4dz, sorry about that :)

m4dz:
My bad! Indeed, you can test w/ a `watch:standalone` task in background, an pointung your browser on `http://localhost:8082`. You also need to deploy some apps in your running stack to display content in apps switcher (see https://trello.com/c/EupO9HTM/198-2-bar-3-e-20-etu-je-peux-naviguer-parmis-mes-applications#comment-5894706c71bf37c33bf52da4)

gregorylegarec:
Great @m4dz, thanks ! 👍

PR #15, opened 2017-02-21
Fix the cozy-bar

nono:


m4dz:
Doesn't it duplicates the `try...catch` matches block?

nono:
@m4dz the selector was wrong in your code. And I don't have `document.matches` in firefox (and probably other browsers too, I didn't check). And if I use `element.matches`, it returns false, it does not throw an exception.

m4dz:
Hmm, al'right, my bad, my code wa clearly wrong, I don't understand how I had tested it :cry:…

Doesn't throwing an exception a better way to handle the lack of given element ?

nono:
> Doesn't throwing an exception a better way to handle the lack of given element ?

The important thing is to print a message in the console. An `if` or a `try / catch` is equally relevant and does the same.

m4dz:
Al'right, let's do this. Thanks for te fix :+1:

PR #21, opened 2017-02-28
Add settings links

m4dz:


src/lib/stack.js
```diff
@@ -30,6 +30,9 @@ module.exports = {
         throw new UnavailableStackException()
       }
     },
+    async settingsBaseURI () {
+      return (await this.apps()).find(item => item.attributes.slug === 'settings').links.related
```
nono: What happen if the settings app is not installed?

m4dz: I though that Settings is always installed, but we can address this use-case

src/components/Bar.svelte
```diff
@@ -51,6 +51,13 @@
     this.set({ config: Object.assign({}, this.get('config'), {apps}) })
   }
 
+  async function updateSettingsURIs () {
+    const baseURI = await stack.get.settingsBaseURI()
+
+    // Assume all Settings' links are all located in first index
```
CPatchane: Maybe put add comment in the yaml file to be aware about this constraint?

m4dz: Since it's expressed in the mockups, it's probably not mandatory. But I agree, we may find a way to be more robust to prevent any code-break… I'll think about it and make some update :+1:

m4dz:
Al'right! This brand-new version takes asynchronously care of the status of the `settings` app, and update the URL of settings entries with a correct baseURI. It also improve the way the config is served, creating internal references in the config tree to have a more robust way to access / update config content.

@GoOz : https://github.com/cozy/cozy-bar/pull/21/commits/3c152df24af612a9831520fabac20bf0e1241bb6 remove bold style on sections items

@y-lohse : https://github.com/cozy/cozy-bar/pull/21/commits/537cf9d24462ec0c00be9807e1b18818807537a1 add a `data-drawer-visible` attribute on the root node (`[role=banner]`) that reflects the status of the cozy-bar drawer

Thanks to all for your patience :smile:

GoOz:
@m4dz thanks 👍

src/components/NavigationSection.svelte
```diff
@@ -1,18 +1,20 @@
 <li class='coz-nav-section' on:click='dispatch(event)'>
   <a on:click='toggle()' aria-controls='{{`coz-nav-pop--${hash}`}}' aria-busy='{{busy}}' data-icon='{{icon}}'>
-    {{t(label)}}
+    {{label}}
   </a>
-  {{#if items}}
+  {{#if items && items.length}}
   <div class='{{`coz-nav-pop coz-nav-pop--${hash}`}}' id='{{`coz-nav-pop--${hash}`}}' aria-hidden={{closed}}>
-    {{#if items.error}}
+  {{#if items[0].error}}
```
CPatchane: Maybe we can prefer `#if items.length && items[0].error` here to prevent some errors. What do you think?

src/lib/stack.js
```diff
@@ -13,31 +16,56 @@ const fetchOptions = {
...
+
+async function hasApp (slug) {
+  return !!(await getApp(slug))
+}
+
 module.exports = {
   init ({cozyURL}) {
     COZY_URL = `//${cozyURL}`
   },
+  has: {
```
CPatchane: Cool 👍

m4dz:
> The only odd thing I noticed is that when I click on "apps", nothing happens. I don't have any other apps installed, so perhaps it's the expected behavior in this situation.

Yup, I noticed it too, but as @kelukelu said, we need at least one app (Files?) installed to have a Cozy working properly ; so we choose to not handle this use-case :smiley:

src/components/NavigationSection.svelte
m4dz: Hmm, I think this ine is useless as we ensure to always pass an Array instance as items, so I'm not sure we need an extra-test :smiley:

CPatchane:
Thanks @m4dz 👍

PR #23, opened 2017-03-07
Logout

m4dz:
This PR enables the logout feature in the settings menu

### DIFF
--- a/src/lib/stack.js
+++ b/src/lib/stack.js
@@ -5,36 +5,46 @@ import 'babel-polyfill'
 
 import {
   UnavailableStackException,
-  UnavailableSettingsException
+  UnavailableSettingsException,
+  UnauthorizedStackException
 } from './exceptions'
 
 // the option credentials:include tells fetch to include the cookies in the
 // request even for cross-origin requests
-const fetchOptions = {
-  credentials: 'include'
+function fetchOptions () {
+  return {
+    credentials: 'include',
+    headers: {
+      Authorization: `Bearer ${COZY_TOKEN}`
+    }
+  }
 }
 
 let COZY_URL = __SERVER__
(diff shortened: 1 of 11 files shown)


src/lib/stack.js
```diff
@@ -5,52 +5,78 @@ import 'babel-polyfill'
...
+
+      try {
+        hasSettings = await hasApp('settings')
+      } catch (e) {
+        hasSettings = false
+        throw new UnavailableSettingsException()
+      }
+
+      if (!hasSettings) {
+        throw new UnavailableSettingsException()
```
CPatchane: Why throwing an error here? I think it should be considered as a normal behaviour that hasSettings is false, should it? It's expected as a function name `hasSomething` to respond `true` or `false` without error if `false` as it's a function to check an existence.

m4dz: I agree w/ your point, but as the Settings app is crucial (w/o it, no possibilities to update password nor change language) so I though throwing an Exception (as well as returning the status of the app, falsy) can be a pretty solution to interrupt Settings links rendering in view and display a wrning message in console about the disabling of those features. But I could be wrong…

m4dz:
@CPatchane did you really review this PR when I pushed it at 01:00 am :fearful: ?

src/lib/stack.js
CPatchane: Good point, I agree about this specific case of the settings app, maybe you can add a comment to be sure to not generalise this way if a similar function is necessary for another app? What do you think?

CPatchane:
Great @m4dz thanks!

src/lib/stack.js
m4dz: Comment added :taco:

PR #24, opened 2017-03-08
[Fix] Fix layout so it won't break on Safari

GoOz:
That's only a part of the fix.


GoOz:
Related : 
https://github.com/cozy/cozy-ui/pull/74

m4dz:
Thanks :rocket:

PR #25, opened 2017-03-08
Improve warning when required markup is missing

clochix:


src/index.js
```diff
@@ -10,6 +10,7 @@ import stack from './lib/stack'
 import BarView from './components/Bar'
 
 const APP_SELECTOR = '[role=application]'
+const BANNER_SELECTOR = '[role=banner]'
```
m4dz: I'm not sure it's really useful as the `[role=banner]` element is the bar itself. When we need to rely on it, we assume the bar is present as we previously injected it, nope?

src/index.js
```diff
@@ -28,7 +29,8 @@ const injectDOM = function CozyBarInjectDOM (data) {
   const barNode = createElement()
   const appNode = document.querySelector(APP_SELECTOR)
   if (!appNode) {
-    return console.warn(`Cozy-bar is looking for a "${APP_SELECTOR}" tag that contains your application and can't find it :'(… The BAR is now disabled`)
+    console.warn(`Cozy-bar is looking for a "${APP_SELECTOR}" tag that contains your application and can't find it :'(… The BAR is now disabled`)
+    return null
```
m4dz: :+1:

src/index.js
```diff
@@ -44,7 +46,12 @@ const bindEvents = function CozyBarBindEvents () {
   document.body.addEventListener('click', this._clickOutsideListener)
 
   this._drawerObserver = this.observe('drawerVisible', drawerVisible => {
-    document.querySelector('[role=banner]').dataset.drawerVisible = drawerVisible
+    const bannerNode = document.querySelector(BANNER_SELECTOR)
+    if (!bannerNode) {
+      console.warn(`Cozy-bar is looking for a "${BANNER_SELECTOR}" tag that contains your banner and can't find it :'(…`)
```
m4dz: This observer is bound only if bar iew is created (see https://github.com/cozy/cozy-bar/pull/25/files#diff-1fdf421c05c1140f6d71444ea2b27638R106), so this condition is probably useless

src/index.js
```diff
@@ -54,11 +61,21 @@ const unbindEvents = function CozyBarUnbindEvents () {
 }
 
 const getDefaultStackURL = function GetDefaultCozyURL () {
-  return document.querySelector('[role=application]').dataset.cozyDomain
+  const appNode = document.querySelector(APP_SELECTOR)
+  if (!appNode) {
+    console.warn(`Cozy-bar is looking for a "${APP_SELECTOR}" tag that contains your application and can't find it :'(… The BAR is now disabled`)
```
m4dz: Great, can we adapt the warning message to reflect the status? In this context, the bar will try to be injected, even if it probably fails at init if the cozyURL isn't supplied explicitely. I suggest something like: "Cozy-bar can't discover the cozy's URL, and will probably fail to initialize the connection with the stack"

src/index.js
```diff
@@ -54,11 +61,21 @@ const unbindEvents = function CozyBarUnbindEvents () {
...
+    return ''
+  }
+  return appNode.dataset.cozyDomain
 }
 
 const getDefaultToken = function GetDefaultToken () {
-  return document.querySelector('[role=application]').dataset.cozyToken
+  const appNode = document.querySelector(APP_SELECTOR)
+  if (!appNode) {
+    console.warn(`Cozy-bar is looking for a "${APP_SELECTOR}" tag that contains your application and can't find it :'(… The BAR is now disabled`)
```
m4dz: Like previous comment, we may adapt the warning message to "Cozy-bar can't discover the app's token, and will probably fail to initialize the connection with the stack"

src/index.js
clochix: Yes, you're definitively right. I hit an issue with a missing element, so I added the test on every element, before understanding that the banner was injected by our code.

clochix:
Thanks for the review, it should be fixed now.

m4dz:
Oups my ba, I totally miss your changes, I'm sorry… Thanks!

PR #26, opened 2017-03-09
Fix perms on icons

m4dz:
Asynchronously loaded base64 encoded blob URI #bullshitBingoYeah \o/


CPatchane:
Great 👍 Thanks @m4dz !

PR #32, opened 2017-04-11
feat: Add request to disk quota

GoOz:
Basically I just copied and pasted what existed before for disk usage. I tried to do both with a one shot request but couldn't get this thing to work with an object instead of a single string.

Also, i didn't copied the part where the data self update, considering that a quota on the disk isn't something that changes frequently and that only one request on the loading of the app is enough.

To test this you should use an updated version of Cozy-stack and launch it manually. Once you're done just use the following commands:
- Use your local (updated) repo of Files (aka Drive) with a `yarn link cozy-bar` in it.
- Then start you stack server with an appropriate path to your local Files' repo:
`cozy-stack serve --appdir drive:$HOME/Repos/cozy-files-v3/build/` 
- Add your instance with a defined quota:
`cozy-stack instances add --dev --disk-quota 50000000000 --passphrase cozy cozy.tools:8080`

**NB: I realized that the stack has an option to say that a cozy instance has a limited quota or not. This PR only takes care of the case with limited quota. If your instance doesn't have a limited quota, you'll have a default quota of 100GB.**

### DIFF
--- a/src/lib/stack.js
+++ b/src/lib/stack.js
@@ -61,7 +61,13 @@ function getDiskQuota () {
 
     return res.json()
   })
-  .then(json => parseInt(json.data.attributes.quota, 10))
+  .then(json => {
+    if (Number.isInteger(json.data.attributes.quota)) {
+      return parseInt(json.data.attributes.quota, 10)
+    } else {
+      return 100000000000 // @TODO Waiting for instructions about how to deal with limitless instances
+    }
+  })
   .catch(e => {
     throw new UnavailableStackException()
   })
--- a/src/lib/stack.js
+++ b/src/lib/stack.js
@@ -52,6 +52,21 @@ function getDiskUsage () {
   })
 }
 
+function getDiskQuota () {
+  return fetch(`${COZY_URL}/settings/disk-usage`, fetchOptions())
+  .then(res => {
+    if (res.status === 401) {
+      throw new UnauthorizedStackException()
+    }
+
+    return res.json()
+  })
+  .then(json => parseInt(json.data.attributes.quota, 10))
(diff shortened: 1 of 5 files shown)


GoOz:
[(1) [BAR-12 / E-36] ETU je peux visualiser le quota de mon disque](https://trello.com/c/W3Quk26R/304-1-bar-12-e-36-etu-je-peux-visualiser-le-quota-de-mon-disque)

src/lib/stack.js
```diff
@@ -61,7 +61,13 @@ function getDiskQuota () {
 
     return res.json()
   })
-  .then(json => parseInt(json.data.attributes.quota, 10))
+  .then(json => {
+    if (Number.isInteger(json.data.attributes.quota)) {
+      return parseInt(json.data.attributes.quota, 10)
+    } else {
+      return 100000000000 // @TODO Waiting for instructions about how to deal with limitless instances
```
gregorylegarec: To clarify that value, could you just use instead declared at the top of the file, with an explicit name, for example
```js
const DEFAULT_DISK_QUOTA = 100000000000
```

GoOz: I get your point but is it really relevant giving the fact that there may not be any default value for limitless instance? This chunk of code is only there to avoid breaking the view, besides that it's totally trashable.
Not that i don't wanna do it, it's just that i don't understand the need for clean trash code.

gregorylegarec: You're right,  I missed your comment with the @TODO. It will fit well until new instructions about how to handle limitless instances. Thanks !

gregorylegarec:
Thanks @GoOz  !

PR #33, opened 2017-04-12
feat: Add hover effect on pop menu links + clean up

GoOz:
- [x] Hover effects on links & buttons in Settings menu
- [x] Hover effects on Apps links
- [x] Cursor pointer only on clickable elements
- [x] Remove `text-transform: capitalize` from links

### DIFF
--- a/src/styles/navigation_item.css
+++ b/src/styles/navigation_item.css
@@ -1,4 +1,6 @@
 [role=banner] .coz-nav-item [role=menuitem] {
+  position: relative;
+  z-index: 0;
   display: block;
   box-sizing: border-box;
   width: 100%;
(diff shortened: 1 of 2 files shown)


GoOz:
[(1) [BAR-13 / E-36] Cosmétique : j'ai une indication lorsque je passe au dessus des élements cliquable](https://trello.com/c/gWrLMX1O/303-1-bar-13-e-36-cosmetique-j-ai-une-indication-lorsque-je-passe-au-dessus-des-elements-cliquable)

src/styles/navigation_item.css
```diff
@@ -11,11 +13,33 @@
...
 }
 
+/* Force pointer on clickable elements*/
+[role=banner] .coz-nav-item a[role=menuitem],
+[role=banner] .coz-nav-item button[role=menuitem] {
+  cursor: pointer;
+}
+
+/* Hover effect */
+[role=banner] .coz-nav-item a[role=menuitem]:hover:before,
```
gregorylegarec: Does we need to specify `.coz-nav-item` here ?

GoOz: Not sure, it could work without it for now I guess, but the original selector for this kind of element is `[role=banner] .coz-nav-item [role=menuitem]` and I guess @m4dz added the class selector to give some higher specificity. 
Because the attribute selector is very low on specificity, those elements could inherit some style that this declaration wouldn't overwrite.

So basically, today it seems to be ok to remove it but it's not future proof. Don't forget that there isn't CSS modules on the Cozy-bar and therefore no scoping at all.
So I wouldn't remove it.

gregorylegarec: Ok for me, let's keep it, @m4dz should had good reasons 👍

gregorylegarec:
Thanks @GoOz !

PR #34, opened 2017-04-12
fix: remove features on mobile

enguerran:
We get errors when cozy-bar setups two features on mobile phone, specially visible when running on offline mode:

- disk-usage statistics
- other apps access

Best solution found: don't setup those features when `target === mobile`. Is this cool?


kosssi:
👍

PR #38, opened 2017-04-17
feat: Remove unnecessary request on mobile :fire:

kosssi:
⚠️ I haven't understanding all cozy-bar, maybe @m4dz can add documentation. So this PR is probably not correct.

This PR remove 2 server requests to approve offline mode.

I think we must more define with @jsilvestre, talk what should do the cozy bar in mobile mode.


kosssi:
I close this PR, you can see a better solution https://github.com/cozy/cozy-bar/pull/34

PR #39, opened 2017-04-25
Force cozy bar to rerender when locale change

CPatchane:
Expose setLocale to change the cozy bar locale from the application

### DIFF
--- a/src/components/Bar.svelte
+++ b/src/components/Bar.svelte
@@ -4,7 +4,7 @@
 </button>
 {{/if}}
 
-<h1 data-lang='{{lang}}' class='{{titleClass}}'>
+<h1 lang='{{lang}}' class='{{titleClass}}'>
   <img class='coz-bar-hide-sm' src='{{iconPath}}' width='32' />
   {{#if appEditor}}<span class='coz-bar-hide-sm'>{{appEditor}} </span>{{/if}}
   <strong>{{appName}}</strong>
--- a/src/components/Bar.svelte
+++ b/src/components/Bar.svelte
@@ -4,7 +4,7 @@
 </button>
 {{/if}}
 
-<h1 class='{{titleClass}}'>
+<h1 data-lang='{{lang}}' class='{{titleClass}}'>
   <img class='coz-bar-hide-sm' src='{{iconPath}}' width='32' />
   {{#if appEditor}}<span class='coz-bar-hide-sm'>{{appEditor}} </span>{{/if}}
   <strong>{{appName}}</strong>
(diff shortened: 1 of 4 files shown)


src/index.js
```diff
@@ -46,6 +46,11 @@ const bindEvents = function CozyBarBindEvents () {
   this._clickOutsideListener = () => this.fire('clickOutside')
   body.addEventListener('click', this._clickOutsideListener)
 
+  document.getElementById('coz-bar').addEventListener('localechange', () => {
+    console.log(this)
```
gregorylegarec: ⚠️  `console.log`

src/index.js
```diff
@@ -138,12 +145,19 @@ const init = function CozyBarInit ({
...
 
   if (view) {
     bindEvents.call(view)
     view.on('teardown', unbindEvents.bind(view))
   }
 }
 
-module.exports = { init, version: __VERSION__ }
+// set the cozy bar locale from the application
+const setLocale = function SetLocale (lang) {
```
gregorylegarec: Is `SetLocale` mandatory between `function` and `(lang)` ?

src/lib/i18n.js
```diff
@@ -20,8 +20,18 @@ const init = function I18nInit (lang) {
   }
 }
 
+const setLang = function I18nSet (lang) {
```
gregorylegarec: Why `setLang` and not `setLocale` ?

src/index.js
CPatchane: Oops! It's fixed^^

src/lib/i18n.js
CPatchane: It was due to the `setLocale` already used in `index.js`, but I renamed it `i18nSetLocale` for better understanding here.

src/index.js
CPatchane: I really don't know for now but all functions and methods are declared in this way in the cozy bar code so I just followed to be coherent.

src/components/Bar.svelte
```diff
@@ -4,7 +4,7 @@
 </button>
 {{/if}}
 
-<h1 class='{{titleClass}}'>
+<h1 data-lang='{{lang}}' class='{{titleClass}}'>
```
lespacedunmatin: Why `data-lang` ? You can also use `lang` directly (is useful for screen readers).

src/index.js
gregorylegarec: So, let's be coherent and keep it as it for now.

src/components/Bar.svelte
CPatchane: In fact I added this attribute just to use the lang data property in the component template to get the correct rerendering behaviour when the lang change. I don't have opinions about that, just I know data-lang is also accepted as attribute. So let's do it 👍

gregorylegarec:
Thanks @CPatchane

src/index.js
m4dz: Having functions stored in variables allow JS engine to evaluate them lazily, which is better for perfomances, but naming functions even in this case make debug messages more explicits :smiley:

PR #43, opened 2017-05-09
chore: Remove useless layout style

GoOz:
I had to moved that style to Cozy-UI because on the mobile app it triggers some kind of ugly repaint where the nav briefly appears on top of the viewport before going down at its rightful place.

Moving that style to UI don't cause the repaint because there's no overwrite anymore so it kinda fix this bug. Furthermore, as the layout evolved I don't think it's needed anymore to have this overwrite of the layout style from the cozy-bar.
Loading an app without the cozy-bar doesn't cause any issue and with cozy-bar we don't have the repaint anymore.

What do you think @m4dz?

### DIFF
--- a/src/styles/base.css
+++ b/src/styles/base.css
@@ -1,25 +1,3 @@
-html {
-  height: 100%;
-}
-
-body {
-  display: flex;
-  flex-direction: column;
-  align-items: stretch;
-  width: 100vw;
-  height: 100%;
-  overflow: hidden;
-}
-
-
-[role=application] {
-  display: flex;
-  height: inherit;
-  flex: 1 1 100%;
-  overflow-y: auto;
-}
-
-
 [role=banner] .coz-sep-flex {
   margin: 0;
   border: none;


GoOz:
[[FIL-46] Corriger l'affichage du chargement de l'app](https://trello.com/c/xKRCKy24/278-fil-46-corriger-l-affichage-du-chargement-de-l-app)

GoOz:
Or maybe it's a reflow and not a repaint issue… I always mixed them up. Either way, it's an issue 😄

GoOz:
@m4dz Now that you say it… it won't work if somebody if using the cozy-bar and not cozy-ui. Do we have that use-case? I might as well let this in cozy-bar's CSS and still add it in cozy-ui but we would have duplicated code… 😕

m4dz:
@GoOz internally we don't face this use-case. But it may have some contributor's apps that uses the cozy-bar and not cozy-ui. BTW, this is not a point yet, so we probably want to merge your PR and address this use-case later :wink:

GoOz:
Fine by me, thanks for your input.

PR #44, opened 2017-05-09
Better apps navigation menu

CPatchane:
* [x] Add row elements limits in apps list (4 for desktop, 3 for mobile)
* [x] Fix bug about labels in menu + better handling apps name and item slug
* [x] Add pagination using scrolling for apps list in mobile drawer
* [x] Better handling long app name in apps navigation

### DIFF
--- a/src/components/NavigationGroup.svelte
+++ b/src/components/NavigationGroup.svelte
@@ -1,6 +1,6 @@
 {{#if separator == 'top'}}<hr />{{/if}}
 {{#if group.length}}
-<ul class='coz-nav-group'>
+<ul class='{{`coz-nav-group${wrappingClass}`}}'>
   {{#each group as item}}
     <NavigationItem item='{{item}}' />
   {{/each}}
@@ -14,6 +14,12 @@
   export default {
     components: {
       NavigationItem
+    },
+    computed: {
+      wrappingClass: (itemsLimit, group) => {
+        if (!itemsLimit || !group.length) return ''
+        return group.length > itemsLimit ? ' --wrapping' : ''
+      }
     }
   }
 </script>
--- a/src/components/NavigationGroup.svelte
+++ b/src/components/NavigationGroup.svelte
@@ -1,6 +1,6 @@
 {{#if separator == 'top'}}<hr />{{/if}}
 {{#if group.length}}
-<ul class='{{`coz-nav-group${wrappingClass}`}}'>
+<ul class='{{`${(wrapping ? "coz-nav-group coz-nav-group--wrapping" : "coz-nav-group")}`}}'>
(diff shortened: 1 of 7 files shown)


src/styles/drawer.css
```diff
@@ -74,6 +74,25 @@
...
+  overflow: scroll;
+  padding-bottom: 1em;
+}
+
+[role=banner] .coz-drawer--apps .coz-drawer--apps-paginator ul {
+  min-width: 100%;
+}
+
+[role=banner] .coz-drawer--apps ul li {
+  flex-basis: calc(100% / 3);
```
GoOz: You should write `flex: 0 0 calc(100% /3);` 
From what i can recall, every browsers don't have the same default values for `flex-grow` and `flex-shrink`

src/components/NavigationGroup.svelte
```diff
@@ -1,6 +1,6 @@
 {{#if separator == 'top'}}<hr />{{/if}}
 {{#if group.length}}
-<ul class='coz-nav-group'>
+<ul class='{{`coz-nav-group${wrappingClass}`}}'>
```
GoOz: I don't get this, maybe it's just the syntax but won't it write `coz-nav-group` if not wrapping and `coz-nav-group--wrapping` if wrapping ? If so, you'll lose the style for `.coz-nav-group` and why in the CSS it is written just`.--wrapping`? I'm confused. 😕

CPatchane: No there is a space (https://github.com/cozy/cozy-bar/pull/44/files#diff-b97a1bf8e7abb24f9a5b3e2a65fcdf69R21) but I think I can put that space here for better understanding 👍

GoOz: ah yeaaaah ok. In that case, i'd rather see a `coz-nav-group--wrapping` to be more consistent cause i don't see any class in Bar's CSS beginning with `--`, and it's kinda confusing.

gregorylegarec: I agree with @GoOz. I spotted this point but did not mention it. Now that @GoOz has spoken, I can tell that we should better use `coz-nav-group--wrapping` to keep consistency with CozyUI logic.

src/components/NavigationGroup.svelte
```diff
@@ -16,9 +16,9 @@
       NavigationItem
     },
     computed: {
-      wrappingClass: (itemsLimit, group) => {
-        if (!itemsLimit || !group.length) return ''
-        return group.length > itemsLimit ? ' --wrapping' : ''
+      wrapping: (itemsLimit, group) => {
```
gregorylegarec: Cleaner solution IMHO 👍

PR #46, opened 2017-05-10
Fix two current bugs

CPatchane:
* [x] Fix bug about random apps menu opening
* [ ] Fix bug about empty or duplicated settings items (seems to be Firefox only)


m4dz:
@gregorylegarec I'm not surprised as resetting `length` in an array, or remapping a variable pointer to new array does the same thing, except in the former the same memory space is used, so it's more efficient and less memory consumming. But if there's a bug in the former case, it remains in the latter. Can you explain me what happens?

gregorylegarec:
@m4dz, of course, the issue is described here https://forum.cozy.io/t/entree-de-menu-disparu-ou-duplique/4383 and occurs only in Firefox.

m4dz:
As thé settings pane made dôme async calls to populate its entries at each opening, I suggestion de delay it with an async instructions like with apps: https://github.com/cozy/cozy-bar/blob/master/src/config/menu.yaml#L34. It may solve this issue.

m4dz:
@CPatchane I tried to make some tests but I haven't a correct setup right now (I can't have a list of installed apps through the docker stack, I need to run one locally, but it's a little bit late yet). Your patch seems ok to me. I wonder if the drawer works too as it uses the same logics for loading items. Have your tested it too?

Thanks for the good works btw :+1: !

src/components/NavigationSection.svelte
```diff
@@ -74,26 +73,26 @@
       return {
         busy: false,
         closed: true,
-        valve: false
+        isFetching: true
```
m4dz: Does passing it to `true` by default not blocking any opening? (but I think we now haven't any other pane that doesn't need to be delayed)

CPatchane: @m4dz I didn't get any problem in my tests but it's using the dev environment, let's see in the prod, what do you think? Do you still have the issue with the setting panel?

m4dz: Yet, it seems to works, but w/ my current setup I can't install the settings app, so the requests are executed quickly, and it _may_ mask some bugs. But yep, go to test it in prod :+1:

PR #51, opened 2017-05-23
Add apps categorisation

CPatchane:


gregorylegarec:
@CPatchane build is broken, could you have a look ?

CPatchane:
@gregorylegarec It was a wrong manipulation, the good build is green here :)

src/components/NavigationSection.svelte
```diff
@@ -13,7 +13,14 @@
     <NavigationGroup group='{{group}}' separator='bottom' />
     {{/each}}
   {{else}}
-    <NavigationGroup group='{{items}}' itemsLimit={{4}} />
+    {{#if categories}}
+      {{#each categories as category}}
+        <p class='coz-nav-category'>{{t(`Categories.${category.title}`)}}</p>
```
lespacedunmatin: Why a `<p>`and not a `<hx>` ?

src/components/NavigationSection.svelte
```diff
@@ -69,6 +76,22 @@
     }
   }
 
+  function getCategories (items) {
+    const categoriesOject = {}
+    items.forEach(i => {
```
lespacedunmatin: So you use the standard Interger var name as an object ? Isn’t this error prone ?

src/components/NavigationSection.svelte
```diff
@@ -69,6 +76,22 @@
...
+      if (!categoriesOject.hasOwnProperty(i.category)){
+        categoriesOject[i.category] = []
+      }
+      categoriesOject[i.category].push(i)
+    })
+    // since sveltejs doesn't allow to loop through objects
+    const categories = []
+    for (const category in categoriesOject) {
+      categories.push({title: category, items: categoriesOject[category]})
```
lespacedunmatin: So that function does not only return a categories array, it also sorts all items into its category… Perhaps you should indicate that in a comment before the function.

src/components/NavigationSection.svelte
```diff
@@ -69,6 +76,22 @@
     }
   }
 
+  function getCategories (items) {
```
lespacedunmatin: You do not strengthen your code for the `grouped: items => items[0] instanceof Array` case ?

gregorylegarec: I suggest to use `reduce` instead of `forEach`:
```js
function getCategorizedItems (items) {
  const categorizedItemsArray = items.reduce((accumulator, item) => {
    accumulator[item.category] = accumulator[item.category] || []      
    accumulator[item.category].push(item)
    return accumulator
   }, {})

  return Object.keys(categorizedItemsArray).map(key => {
    return {key: key, items: categorizedItemsArray[key]}
  })
}
```

src/lib/config.js
```diff
@@ -32,6 +33,7 @@ async function updateAppsItems (config) {
           slug: app.attributes.slug,
           l10n: false,
           href: app.links.related,
+          category: CATEGORIES.includes(app.attributes.category) ? app.attributes.category : 'others',
```
CPatchane: `babel-polyfill` handle this polyfill ([here](https://github.com/cozy/cozy-bar/blob/master/src/index.js#L5)) and since it was already use ([here for example](https://github.com/cozy/cozy-bar/blob/master/src/lib/config.js#L15)) I guess it's correctly handled 👍

src/locales/en.json
```diff
@@ -10,5 +10,11 @@
   "help": "Help",
   "logout": "Sign out",
   "error_UnavailableStack": "The stack is unreachable (connection timed-out).",
-  "error_UnauthorizedStack": "Some permissions are missing, the application can't access the requested resource on the stack."
+  "error_UnauthorizedStack": "Some permissions are missing, the application can't access the requested resource on the stack.",
+
+  "Categories": {
```
CPatchane: I used a capital for the tree root here like in [photos](https://github.com/cozy/cozy-photos-v3/blob/master/src/locales/en.json) but It was a only a personal preference here.

src/components/NavigationSection.svelte
CPatchane: Good point! I will rename this function to be more explicit. And @gregorylegarec it's `items` instead of `apps` because it's not specific to the apps here, you can use the [`categorized` property](https://github.com/cozy/cozy-bar/pull/51/files#diff-19e2638c4d86ae90b81f1d3f8354850dR36) to categorize other items in general.

gregorylegarec: Ok I think the case will never happen but ok, it is irrelevant to rename all the items.

src/locales/en.json
gregorylegarec: Ok, we just should set up a convention together and stick to it.

src/components/NavigationSection.svelte
CPatchane: Thanks for this improvement, I will try it 👍

CPatchane:
[(5) [BAR-15 / E-36] ETU j'ai une hierarchie dans mes applications](https://trello.com/c/aM0IKdT9/360-5-bar-15-e-36-etu-j-ai-une-hierarchie-dans-mes-applications)

gregorylegarec:
Geat @CPatchane !

PR #53, opened 2017-05-24
Add beta status

CPatchane:
* [x] Drawer
* [x] Settings menu
* [x] Next to the app name

### DIFF
--- a/src/styles/navigation_section.css
+++ b/src/styles/navigation_section.css
@@ -2,7 +2,6 @@
   position: relative;
 }
 
-
 [role=banner] .coz-nav-section [aria-controls] {
   display: flex;
   align-items: baseline;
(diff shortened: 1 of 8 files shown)


CPatchane:
[(3) ⭐️ [DIV-2] ETU j'ai conscience d'être sur une version beta](https://trello.com/c/YXTAPdtU/361-3-div-2-etu-j-ai-conscience-d-etre-sur-une-version-beta)

src/styles/navigation_section.css
```diff
@@ -90,6 +96,10 @@
     padding: 0 .5em;
 }
 
+[role=banner] .coz-nav-group:not(.coz-nav--error).coz-nav-group--grey {
```
gregorylegarec: Is `--grey` the  best semantic specifier ? Maybe something like `--inactive` should be better ? What do you think ?

CPatchane: Indeed, that sounds better for the semantic here, I agree 👍

gregorylegarec:
Great @CPatchane !

PR #60, opened 2017-06-12
Feat: add PTNB app category

gregorylegarec:


src/locales/en.json
```diff
@@ -17,6 +17,7 @@
   "Categories": {
     "cozy": "Cozy apps",
     "partners": "Partners apps",
+    "ptnb": "expPTNB",
```
CPatchane: Are you sure about "expPTNB"? It's not very explicit as a locale here.

gregorylegarec: Yes I am totally sure, this category is aimed to be used only in French. Don't forget to avoid over-quality ;)

CPatchane: Right so 👍

PR #63, opened 2017-06-20
feat: manage Coming Soon Apps from context

gregorylegarec:
My `cozy.yaml` looks like:
```yaml
# It is possible to customize some behaviors of cozy-stack in function of the
# context of an instance (the context field of the settings document of this
# instance). Here, the "beta" context is customized with.
contexts:
  maif_fing:
    # Redirect to a specific route of cozy-collect after the onboarding
    onboarded_redirection: collect/#/discovery/?intro
    # Redirect to the photos application after login
    default_redirection: drive/#/files
    # Allow to customize the cozy-bar link to the help
    help_link: https://forum.cozy.io/
    coming_soon:
      store:
        editor: 'Cozy'
        name: 'Store'
        slug: 'store'
        category: 'cozy'
      bank:
        editor: 'Cozy'
        name: 'Bank'
        slug: 'bank'
        category: 'partners'
      sante:
        editor: 'Cozy'
        name: 'Santé'
        slug: 'sante'
        category: 'partners'
```
I am waiting for @nono to know if it is possible to specify a `default` context, when no `context` property is specified in the instance settings.


gregorylegarec:
[(3) [BAR-19] ETU MesInfos, j'ai l'application bank et sante en coming soon](https://trello.com/c/Gg1mgHSZ/444-3-bar-19-etu-mesinfos-jai-lapplication-bank-et-sante-en-coming-soon)

src/lib/config.js
```diff
@@ -2,15 +2,47 @@ import deepClone from 'deep-clone'
...
+        context.data.attributes['coming_soon'] &&
+          Object.values(context.data.attributes['coming_soon']) || []
+
+      cachedComingSoonApps = comingSoonApps.map(app => {
+        let icon
+
+        try {
+          icon = app.slug && {
+            cached: true,
+            src: require(`../assets/icons/comingsoon/icon-${app.slug.toLowerCase()}.svg`)
```
CPatchane: Why use only lowercases in names here? I would prefer folder named comingSoon and to not use `toLowerCase()` function to keep all coherent since we use camelCase everywhere. What do you think?

gregorylegarec: I think that we use CamelCase only for Component and I hate using CamelCase in files/folder names. 😄

gregorylegarec: @CPatchane What about `assets/icons/apps/` ?

CPatchane: Well, we can keep `comingsoon`, it's not so bad (apps will be worst since we could think that main apps icons are stored here which is wrong). But I am not convinced by the `toLowerCase` necessity for the icon name computing here.

CPatchane: I think we shouldn't change the `slug` if we need to use it. If it's not "formatted" correctly or as we want, that means it's not correct at the source and we should change the source and use it as provided. Otherwise it can be confusing as the slug will be formatted differently from the source in the icon name.

gregorylegarec: You're right.

CPatchane:
Thanks @gregorylegarec 🎉

PR #65, opened 2017-06-20
feat: remove hover for coming soon apps :lipstick:

gregorylegarec:
Removes hover effect and pointer cursor for Coming Soon apps. What do you think ?

comingsoon

### DIFF
--- a/src/styles/navigation_item.css
+++ b/src/styles/navigation_item.css
@@ -23,8 +23,8 @@
 }
 
 /* Hover effect */
-[role=banner] .coz-nav-item a[role=menuitem]:hover:before,
-[role=banner] .coz-nav-item a[role=menuitem]:focus:before,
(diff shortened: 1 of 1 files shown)


src/styles/navigation_item.css
```diff
@@ -23,8 +23,8 @@
 }
 
 /* Hover effect */
-[role=banner] .coz-nav-item a[role=menuitem]:hover:before,
-[role=banner] .coz-nav-item a[role=menuitem]:focus:before,
+[role=banner] .coz-nav-item a[role=menuitem]:not(.coz-bar-coming-soon-app):hover:before,
```
GoOz: This selector is getting more insane every time… Wouldn't be simpler to just… i don't know, not putting the coming soon item inside a `<a>` tag?

gregorylegarec: The code is far too abstract for such specificity, but I will give it a try. 👍

CPatchane:
Thanks @gregorylegarec 👍

PR #66, opened 2017-06-22
Say hi! to Claudy :tada:

CPatchane:
* Add claudy
* Upgrade cozy-ui to beta36
* Add claudy Actions with description and link
* Add Piwik tracker for Claudy (if Tracking preference is true)


src/components/Claudy.svelte
```diff
@@ -0,0 +1,57 @@
...
+      }
+    },
+
+    oncreate () {
+      // if tracking enabled, init the tracker
+      if (shouldEnableTracking()) {
+        const trackerInstance = getTracker(__PIWIK_TRACKER_URL__, __PIWIK_SITEID__, false, false)
+        configureTracker({
+          appDimensionId: __PIWIK_DIMENSION_ID_APP__,
+          app: 'Cozy Bar',
```
y-lohse: Maybe you should check with @jsilvestre under what app name Claudy visits should appear. It's quite possible this should use the name of the actual app and not Cozy Bar, but idk.

src/lib/piwik.js
```diff
@@ -0,0 +1,126 @@
+/* global __PIWIK_TRACKER_URL__ __PIWIK_SITEID__ __PIWIK_DIMENSION_ID_APP__ */
```
y-lohse: I'm guessing there's no wya to import this from cozy-client-js to avoid duplication? In that case did you consider removing the unused parts, like the middleware for example?

src/styles/claudy_menu.css
```diff
@@ -0,0 +1,162 @@
...
+  transition: .2s all ease-in-out;
+}
+
+[role=banner] .coz-claudy-menu--action-selected .coz-claudy-menu-header-back-button {
+  opacity: 1;
+  left: 1em;
+  transition: .2s all ease-in-out;
+}
+
+[role=banner] .coz-claudy-menu--action-selected .coz-claudy-menu-header {
```
y-lohse: Empty selector here ;)

CPatchane: Oops, it's fixed. Thanks 👍

src/lib/piwik.js
CPatchane: We don't have any `.jsx` file in the bar so I didn't want to handle it just for this lib which could be in `.js` in the future. So I wanted to keep the exact same file as in cozy-ui and plan to replace it by the an import from cozy-ui when it will be in `.js` that I think that could be possible.

src/styles/bar.css
```diff
@@ -6,7 +6,7 @@
   display: flex;
   align-items: stretch;
   padding: 0 1.25em 0 1em;
-  box-shadow: inset 0 -1px 0 0 var(--grey-09);
+  box-shadow: inset 0 -1px 0 0 var(--silver);
```
GoOz: 👍

src/styles/claudy.css
```diff
@@ -0,0 +1,34 @@
...
+  right: 2em;
+}
+
+@media (min-width: 64em) {
+  [role=banner] .coz-claudy {
+    bottom: 2em;
+  }
+}
+
+[role=banner] .coz-claudy .coz-claudy-icon {
```
GoOz: Why not just `[role=banner] .coz-claudy-icon` as the selector? 
Unless you really need extra specificity weight, which I doudt, the naming convention should be enough to understand that `coz-claudy-icon` is a child of `coz-claudy` I think.

src/styles/claudy.css
```diff
@@ -0,0 +1,34 @@
...
+  width: 3.5em;
+  height: 3.5em;
+  border-radius: 100%;
+  border: none;
+  background-color: var(--dodger-blue);
+  background-image: url('../assets/icons/16/icon-claudy.svg');
```
GoOz: You could have write the shorthand here, like this:
`background: var(--dodger-blue) url('../assets/icons/16/icon-claudy.svg') .5em / 2.5em no-repeat;`

src/styles/claudy_menu.css
```diff
@@ -0,0 +1,158 @@
+[role=banner] .coz-claudy .coz-claudy-menu {
```
GoOz: Same here about over-specified selector.

src/styles/claudy_menu.css
```diff
@@ -0,0 +1,158 @@
...
+[role=banner] .coz-claudy-menu .coz-claudy-menu-title {
+  opacity: 1;
+  transition: .2s all ease-in-out;
+}
+
+[role=banner] .coz-claudy-menu .coz-claudy-menu-header-back-button {
+  position: absolute;
+  width: 2em;
+  height: 2em;
+  background: url('../assets/icons/24/icon-arrow-left.svg') center no-repeat;
```
GoOz: Could be shortened. Friendly reminder:
`background: color url(image) position / size repeat;`

src/styles/claudy_menu.css
```diff
@@ -0,0 +1,158 @@
...
+
+[role=banner] .coz-claudy-menu--action-selected .coz-claudy-menu-content {
+  transform: translateX(-50%);
+  transition: .2s transform ease-in-out;
+}
+
+[role=banner] .coz-claudy-menu .coz-claudy-menu-actions-list,
+[role=banner] .coz-claudy-menu .coz-claudy-menu-action-description {
+  min-width: 50%;
+  max-width: 50%;
```
GoOz: So `width: 50%;` should be enough or did i miss something?

src/styles/index.styl
```diff
@@ -6,26 +6,45 @@
...
   --z-index-over-selection: $selection-index + 1;
 }
 
 [role=banner] {
-  --grey-08: grey-08;
-  --grey-09: grey-09;
-  --grey-10: grey-10;
-  --grey-11: grey-11;
-  --grey-12: grey-12;
+  --charcoal-grey: charcoal-grey;
```
GoOz: My man! 🥇

src/styles/claudy_menu.css
CPatchane: No, it doesn't do the job here, I have to force the width with these properties :/

src/styles/claudy.css
CPatchane: Thanks!

src/styles/claudy_menu.css
CPatchane: Thanks, I was searching quickly for this shorthand but I didn't find with the background size 👍

src/styles/claudy_menu.css
CPatchane: Finally here, I can keep it since we will have to reuse the `.coz-claudy-menu` in the mobile drawer and the styles will changes.

src/styles/claudy_menu.css
GoOz: oh ok then.

gregorylegarec:
[(8) [NUA-1 / E-] ETU j'ai un nuage qui m'affiche une liste de cas d'usages pour desktop](https://trello.com/c/Ss2Na3jG/404-8-nua-1-e-etu-jai-un-nuage-qui-maffiche-une-liste-de-cas-dusages-pour-desktop)

src/components/ClaudyMenu.svelte
```diff
@@ -54,8 +71,32 @@
...
         selectedAction: null,
+        selectedActionUrl: '',
         getIcon (iconName) {
           return require(`../assets/icons/claudyActions/${iconName}`)
+        },
+        computeUrl (action) {
+          const appsList = this.appsList
+          if (action.link.type === 'apps' && action.link.appSlug) {
+            if (!appsList) {
+              console.warn('No apps found on the Cozy')
```
gregorylegarec: Do we really need a `console.warn` here ?

src/components/Bar.svelte
```diff
@@ -19,6 +19,7 @@
 
 {{#if target !== 'mobile' && !isPublic}}
 <Drawer content='{{config.apps}}' footer='{{config.sections.drawer}}' visible={{drawerVisible}} on:close='toggleDrawer(true)'/>
+<Claudy config='{{clouzyConfig}}' appsList='{{config.apps}}'/>
```
gregorylegarec: s/clouzyConfig/claudyConfig/

gregorylegarec:
Threee more things :
  * My logo is not displayed
  * What about setting `outline: 0` on Claudy ?
  * On Firefox there is a tiny space between the tooltip and its "tail"
capture d ecran 2017-06-22 a 16 39 46
capture d ecran 2017-06-22 a 16 41 37

src/components/ClaudyMenu.svelte
CPatchane: Since the action button can be disabled for this reason I would prefer to warn it here. At least until we have a link to the store to put instead. What do you think?

CPatchane:
Thanks @gregorylegarec 👍 
Now the :focus/:active use the same styles than the :hover, thanks @gregorylegarec.
I fixed the logo display issue but I can not reproduce the last issue :/

gregorylegarec:
@CPatchane It looks great ! ❤️ 

In my developer tools, doing the following fixes the Firefox issue.
```css
.cozy-claudy-menu::after {
  bottom: -.5em;
}
```

gregorylegarec:
Amazing job @CPatchane ! 🎸 🎶

PR #70, opened 2017-06-23
Display Claudy according to the context + minor fixes

CPatchane:
Here is my example of `cozy.yaml` for the stack:
```yaml
# It is possible to customize some behaviors of cozy-stack in function of the
# context of an instance (the context field of the settings document of this
# instance). Here, the "beta" context is customized with.
contexts:
  beta:
    # Redirect to a specific route of cozy-collect after the onboarding
    onboarded_redirection: collect/#/discovery/?intro
    # Redirect to the photos application after login
    default_redirection: drive/#/files
    # Allow to customize the cozy-bar link to the help
    help_link: https://forum.cozy.io/
    # Coming soon applications listed in the Cozy Bar's app panel
    # Will be removed when the store will be available.
    coming_soon:
      store:
        editor: 'Cozy'
        name: 'Store'
        slug: 'store'
        category: 'cozy'
    # Claudy action slugs list
    claudy_actions:
      - desktop
      - cozy-collect
      - support
```

The actions list order will depend of the slugs list `claudy_actions` order of this context. And if the `claudy_actions` property is missing, no Claudy will be displayed.
Each slug must matched a property of the `actions` object declared in the `./config/claudy.yaml` config file (an unrecognised slug will simply be ignored).

### Documentation
* https://github.com/cozy/cozy-stack/pull/762
* [x] Readme 
* [x] Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,19 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Removed
 - none yet
 
+## [v3.1.0] - 2017-06-XX
+### Changed
+- none yet
+
+### Fixed
+- none yet
+
+### Added
+- Claudy: a list of actions available for the Cozy and suggested to the user (according the Cozy context setting)
+
+### Removed
+- none yet
+
 
 ## [v3.0.1] - 2017-06-23
 ### Fixed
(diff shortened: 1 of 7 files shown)


gregorylegarec:
@CPatchane Looks great. If I understand well, we are keeping `src/config/claudy.yaml` as a fallback ?

CPatchane:
@gregorylegarec No, the `claudy.yaml` is where you declare all actions available with their options for Claudy and it's the context which will define which one use in which order.
If there is nothing about claudy_actions in context, Claudy won't be displayed.

gregorylegarec:
Great @CPatchane, as usual.

PR #72, opened 2017-06-26
Put ‘others’ category always at the end of the apps list

CPatchane:


src/lib/helpers.js
```diff
@@ -0,0 +1,23 @@
...
+    return accumulator
+  }, {})
+
+  return Object.keys(categorizedItemsObject)
+    .map(category => {
+      return {title: category, items: categorizedItemsObject[category]}
+    })
+    // categories alphabetical sorting
+    .sort((c1, c2) => {
+      if (c1.title === 'others') return 1
```
CPatchane: The only one changes line for this function is here.

CPatchane:
[(1) [BAR-18] ETU je vois la catégorie 'Autre' dans ma cozy-barre toujours en dernière position](https://trello.com/c/W9Vtgej3/428-1-bar-18-etu-je-vois-la-cat%C3%A9gorie-autre-dans-ma-cozy-barre-toujours-en-derni%C3%A8re-position)

src/lib/helpers.js
```diff
@@ -0,0 +1,23 @@
...
+  if (items[0] instanceof Array) return null // doesn't handle this case
+  const categorizedItemsObject = items.reduce((accumulator, item) => {
+    accumulator[item.category] = accumulator[item.category] || []
+    accumulator[item.category].push(item)
+    return accumulator
+  }, {})
+
+  return Object.keys(categorizedItemsObject)
+    .map(category => {
+      return {title: category, items: categorizedItemsObject[category]}
```
gregorylegarec: Could we use `slug` instead of `title` ? In my opinion `title` is more a "human readable" property, which may be displayed to user, and slug a technical one. What do you think ?

CPatchane: You're right, it's not the subject of this PR (it was already here) but I can change it. Thanks 👍

CPatchane: Fixed.

gregorylegarec:
Thanks @CPatchane !

PR #74, opened 2017-06-26
Display drawer claudy section only if claudy actions in context

CPatchane:


src/components/Bar.svelte
```diff
@@ -72,6 +72,12 @@
         await updateSettings(config)
         await updateApps(config)
       }
+      // filter claudy subsection if claudyConfig is null
+      if (!claudyConfig) {
+        if (config.sections && config.sections.drawer) {
+          config.sections.drawer = config.sections.drawer.filter(s => s.length && s[0].slug !== 'claudy')
```
gregorylegarec: What means `s` ?

CPatchane: Subsections, I used just the initial letter in functions like `sort()` or `filter()` to keep them short if they can be written in one line.

gregorylegarec:
Thanks @CPatchane !

src/components/Bar.svelte
gregorylegarec: Ok, I am not a big fan of _one-letter variables_, but let's keep this one. 😄

PR #78, opened 2017-06-27
Release v3.1.1 :rocket:

CPatchane:
### Fixed
 - Force claudy to rerender when the locale changes
 - Fix bug about others apps category not alwayes at the end of the list
 - Don't allow to close the drawer when Claudy is opened
 - Fix bug when app detection for Claudy app links

 ### Added
 - Piwik environment variables for production build


gregorylegarec:
Thanks @CPatchane !

PR #82, opened 2017-07-06
Refacto: Welcome to preact

CPatchane:
Refactoring all the cozy bar using preact/preact-compat and remove svelteJS usage


CPatchane:
[Passer le cozy-bar sous preact](https://trello.com/c/y8ZdQ0d2/504-passer-le-cozy-bar-sous-preact)

gregorylegarec:
Impressive !

Some feedback:

On a clean environment (no apps, no settings): 
If I click on _Apps_ the following style is applied:
capture d ecran 2017-07-07 a 00 38 13

When I click on _Settings_ a spinner is displayed indefinetly:
capture d ecran 2017-07-07 a 00 40 17
and the following error is thrown:
capture d ecran 2017-07-07 a 00 40 54

config/webpack.config.prod.js
```diff
@@ -15,7 +15,6 @@ module.exports = {
     new webpack.DefinePlugin({
       __SERVER__: false,
```
enguerran: I am not sure it is used anywhere.

src/components/Bar.jsx
```diff
@@ -8,7 +8,7 @@ import { shouldEnableTracking, getTracker, configureTracker } from '../lib/piwik
 // import Apps from './Apps'
 // import Settings from './Settings'
 // import Drawer from './Drawer'
-// import Claudy from './Claudy'
+import Claudy from './Claudy'
```
enguerran: At the end, the comments are uncommented in further commits. Hard to review commit by commit this way.

src/components/Bar.jsx
```diff
@@ -69,22 +71,23 @@ class Bar extends Component {
...
+            <button class='coz-bar-burger' onClick={this.toggleDrawer} data-icon='icon-hamburger'>
+              <span class='coz-bar-hidden'>{t('menu')}</span>
+            </button>
+            {claudyConfig &&
+              <Claudy config={claudyConfig} usageTracker={usageTracker} onToggle={this.toggleClaudy} opened={claudyOpened} />
+            }
+          </div>
+        }
       </div>
     )
```
enguerran: I see these variables are used in further commits, so I review a out-of-date commit. Hard to review with chopped commits.

src/components/ClaudyMenu.jsx
```diff
@@ -0,0 +1,149 @@
...
+                  }
+                </div>
+              </div>
+            }
+          </div>
+        </div>
+      </div>
+    )
+  }
+}
```
enguerran: That is a very big render function. Can't we split it in smaller functional component with SRP on each?

src/components/Bar.jsx
```diff
@@ -11,10 +11,11 @@ import { shouldEnableTracking, getTracker, configureTracker } from 'cozy-ui/reac
 import Claudy from './Claudy'
 
 class Bar extends Component {
-  constructor (props) {
+  constructor (props, context) {
     super(props)
+    this.store = context.store
     this.state = {
-      claudyConfig: null, // no claudy by default
+      claudyActions: null, // no claudy by defaul
```
enguerran: s/defaul/default/

(in any case I guess `claudyActions: null` has got enough meaning and the comment is useless)

src/components/Bar.jsx
```diff
@@ -23,6 +24,11 @@ class Bar extends Component {
     this.toggleClaudy = this.toggleClaudy.bind(this)
   }
 
+  async componentWillMount () {
+    const claudyActions = await this.store.getClaudyActions()
+    this.setState({ appsList, claudyActions })
+  }
```
enguerran: I am asking myself if `componentWillMount` is a good place to make asynchronous calls.
In [the documentation](https://facebook.github.io/react/docs/react-component.html#componentwillmount) there is nothing about it. But I see something in previous documentation referenced [in this blog](https://devbutze.com/p/131).

src/lib/BarStore.js
```diff
@@ -0,0 +1,55 @@
...
+      .then(context => {
+        const contextActions = (context.data && context.data.attributes && context.data.attributes['claudy_actions']) || null
+        if (!contextActions) return null
+        // get an arrays of action
+        const claudyActions = contextActions.map(slug => {
+          if (CLAUDY_ACTIONS.hasOwnProperty(slug)) {
+            // adding also the action slug
+            return Object.assign({}, CLAUDY_ACTIONS[slug], { slug })
+          }
+        }).filter(action => action)
```
enguerran: I cannot understand this `filter`.

src/lib/BarStore.js
```diff
@@ -0,0 +1,55 @@
...
+        this.claudyActions = claudyActions
+        return {
+          actions: claudyActions
+        }
+      })
+      .catch(error => {
+        console.warn && console.warn(`Cozy-bar cannot fetch Claudy: ${error.message}`)
+        return null
```
enguerran: Not sure that it is needed to `return null` in a `catch` block.

src/lib/BarStore.js
```diff
@@ -0,0 +1,55 @@
...
+
+  constructor (props, context) {
+    super(props, context)
+    this.store = props.store
+  }
+
+  render ({children}) {
+    return (children && children[0]) || null
+  }
+}
```
enguerran: This `class` is very mysterious. I understand the need is to pass to this components `props` that will be inject in `childContext` for `children`. Why do you instantiate `this.store` instead of returning `getChildContext () { return { store: this.props.store } }`?

src/lib/BarStore.js
```diff
@@ -12,31 +12,119 @@ const CATEGORIES = ['cozy', 'partners', 'ptnb']
...
-        if (!contextActions) return null
-        // get an arrays of action
-        const claudyActions = contextActions.map(slug => {
-          if (CLAUDY_ACTIONS.hasOwnProperty(slug)) {
-            // adding also the action slug
-            return Object.assign({}, CLAUDY_ACTIONS[slug], { slug })
+  async fetchApps () {
+    let apps
+    try {
+      apps = await Promise.all((await stack.get.apps())
```
enguerran: Why do we need to wrap `await stack.get.apps()` in a `Promise.all` functions?

src/lib/BarStore.js
```diff
@@ -12,31 +12,119 @@ const CATEGORIES = ['cozy', 'partners', 'ptnb']
...
+    })
+  }
+
+  async getAppsList () {
+    const apps = await this.fetchApps()
+    const comingSoonApps = await this.fetchComingSoonApps()
+    return apps.concat(comingSoonApps)
+  }
+
+  getClaudyActions () {
```
enguerran: You mix `async/await` syntax and `Promise.then` syntax. It is a little bit hard to read heterogeneous syntax. If there is no special unknown needs, could we homogenize them?

src/lib/BarStore.js
```diff
@@ -12,31 +12,119 @@ const CATEGORIES = ['cozy', 'partners', 'ptnb']
...
+    .then(context => {
+      const contextActions = (context.data && context.data.attributes && context.data.attributes['claudy_actions']) || null
+      if (!contextActions) return null
+      // get an arrays of action
+      const claudyActions = contextActions.map(slug => {
+        if (CLAUDY_ACTIONS.hasOwnProperty(slug)) {
+          // adding also the action slug
+          return Object.assign({}, CLAUDY_ACTIONS[slug], { slug })
+        }
+      }).filter(action => action)
```
enguerran: I already saw this instruction. Don't we need to refactor it?

src/components/Nav.jsx
```diff
@@ -0,0 +1,48 @@
+import React, { Component } from 'react'
+
+import { translate } from 'cozy-ui/react/I18n'
+
+class Nav extends Component {
+  constructor (props, context) {
+    super(props)
+    this.store = context.store
```
enguerran: Why is it used for?

src/components/AppsList.jsx
```diff
@@ -0,0 +1,51 @@
...
+                </a>
+              </li>
+            })}
+          </ul>
+          <hr />
+        </div>
+      )
+    })}
+  </div>
+)
```
enguerran: Very big, could we improve readability (for future maintenance)?

src/components/Bar.jsx
```diff
@@ -27,9 +26,9 @@ class Bar extends Component {
   }
 
   async componentWillMount () {
-    const appsList = await this.store.getAppsList()
+    await this.store.fetchAppsList()
```
enguerran: How is it used? What do happened if we remove this _side-effect_ instruction?

src/lib/BarStore.js
```diff
@@ -124,9 +125,9 @@ export default class BarStore {
       })
   }
 
-  getStorageStats () {
-    return stack.get.storageStats()
-      .then(statsObject => statsObject)
+  getStorageData () {
+    return stack.get.storageData()
+      .then(dataObject => dataObject)
```
enguerran: What is the need of this identity function?

src/lib/stack.js
```diff
@@ -58,12 +58,13 @@ function fetchJSON (url, options) {
...
 
-function getStorageStats () {
+function getStorageData () {
   return fetchJSON(`${COZY_URL}/settings/disk-usage`, fetchOptions())
   .then(json => {
     return {
       usage: parseInt(json.data.attributes.used, 10),
-      quota: parseInt(json.data.attributes.quota, 10),
+      // TODO Better handling when no quota provided
+      quota: parseInt(json.data.attributes.quota, 10) || 100000000000,
```
enguerran: A lot of magic numbers in this file.

src/lib/BarStore.js
```diff
@@ -87,7 +87,11 @@ export default class BarStore {
 
   async fetchAppsList () {
     const apps = await this.fetchApps()
-    const comingSoonApps = await this.fetchComingSoonApps()
+    let comingSoonApps = await this.fetchComingSoonApps()
+    comingSoonApps = comingSoonApps
+      // drop coming soon apps already installed
+      .filter(comingSoonApp =>
+        apps.filter(app => app.slug === comingSoonApp.slug))
```
enguerran:
```js
const comingSoonApps = await this.fetchComingSoonApps()
    .filter(comingSoonApp => apps.filter(app => app.slug === comingSoonApp.slug))
```

src/lib/BarStore.js
```diff
@@ -87,7 +87,11 @@ export default class BarStore {
...
     this.appsList = apps.concat(comingSoonApps)
```
enguerran: side-effects are very dangerous 😨

src/components/Bar.jsx
```diff
@@ -46,8 +46,9 @@ class Bar extends Component {
   toggleDrawer () {
     // don't allow to toggle the drawer if claudy opened
     if (this.state.claudyOpened) return
-    this.props.onDrawer(!this.state.drawerVisible)
-    this.setState({ drawerVisible: !this.state.drawerVisible })
+    const drawerVisible = !this.state.drawerVisible
+    this.props.onDrawer(drawerVisible)
+    this.setState({ drawerVisible })
```
enguerran: Not sure it helps readability.

src/lib/I18n.js
```diff
@@ -0,0 +1,89 @@
+/**
```
enguerran: I though we have to use the I18N from cozy-ui, haven't we?

package.json
```diff
@@ -46,21 +50,27 @@
...
+    "date-fns": "^1.28.5",
     "node-polyglot": "^2.2.2",
     "piwik-react-router": "^0.8.2",
     "preact": "^8.1.0",
     "preact-compat": "^3.16.0"
+  },
+  "standard": {
+    "env": [
+      "browser",
+      "jest"
```
CPatchane: That is just a copy/paste of the config, `jest` will be added later for the tests

src/components/Bar.jsx
CPatchane: Sorry for that, but these comments are correctly removed :)

src/components/Bar.jsx
CPatchane: What do you mean? I need to run a parent method on drawer toggling here

src/components/Bar.jsx
CPatchane: It just won't fetch the appsList in the store and you will have an empty apps list.

package.json
enguerran: Maybe later we will add another test runner. Maybe later we will not remember why there is such a configuration in this very file. The butterfly effect and the chaos theory warn us not to expect what the future will be and we have enough to do to manage present and needed things.

But it is a detail, and noone cares for such little not annoying things but me, too perfectionist. Sorry if it bothers you.

CPatchane: Agree with you but I planned to add tests in the cozy-bar after that refactoring (as preact helps a lot for that) as soon as possible

src/index.jsx
```diff
@@ -0,0 +1,108 @@
+/* global __TARGET__, __VERSION__, __DEVELOPMENT__ */
+
+'use strict'
+
+import 'babel-polyfill'
```
CPatchane: We need to import it for webpack (a require at the root file).

src/index.jsx
```diff
@@ -0,0 +1,108 @@
...
+  const appNode = document.querySelector(APP_SELECTOR)
+  return appNode.dataset.cozyEditor || undefined
+}
+
+const getDefaultIcon = () => {
+  const linkNode = document.querySelector('link[rel="icon"][sizes^="32"]')
+  if (linkNode !== null) {
+    return linkNode.getAttribute('href')
+  } else {
+    return 'data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7'
```
CPatchane: That's an icon encoded in Base64, since it' short to write and used only here, we could keep it.

src/components/ClaudyMenu.jsx
CPatchane: Since we won't reuse theses smaller components if I split it except for here, I would prefer to keep all here, but that's my personal opinion.

src/lib/BarStore.js
CPatchane: That's just doesn't work, I already tried it :/

src/lib/BarStore.js
CPatchane: Removed 👍

src/lib/BarStore.js
CPatchane: This is to prevent having `undefined` or `null` in the actions array.

src/lib/I18n.js
CPatchane: Yes I started like that but I came back because `cozy-ui` needs a dependency to `date-fns` which is just 80kb (production build) of useless code for the cozy-bar here. So this is all the code from cozy-ui without the `date-fns` and `format` parts.

src/lib/BarStore.js
CPatchane: Hum, it was already here, without that it seems to break so I will investigate this later, not in this PR.

src/components/Bar.jsx
```diff
@@ -0,0 +1,90 @@
...
+      })
+      trackerInstance.push(['disableHeartBeatTimer']) // undocumented, see https://github.com/piwik/piwik/blob/3.x-dev/js/piwik.js#L6544
+      this.set({usageTracker: trackerInstance})
+    }
+  }
+
+  toggleDrawer () {
+    // don't allow to toggle the drawer if claudy opened
+    if (this.state.claudyOpened) return
+    this.setState({ drawerVisible: !this.state.drawerVisible })
```
CPatchane: Good point, thanks for this ressource 👍 but I think it's too far for now in this PR.

src/lib/BarStore.js
CPatchane: As these parts will be handled by the inter-app very soon I suggest to keep it like this since it works. What do you think?

package.json
gregorylegarec: I agree with @enguerran, we should find in code only what we are actually using.

src/index.jsx
gregorylegarec: I think @enguerran was suggesting to add a comment, or better IMHO, use a variable:
```js
} else {
  const defaultBase64Icon = 'data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7'
  return defaultBase64Icon.
```
This icon could be declared globally like some kind of config variable.

src/components/ClaudyMenu.jsx
gregorylegarec: I am #TeamSmallComponent but I also trust @CPatchane, I am lost ...

src/lib/BarStore.js
gregorylegarec: @enguerran Actually this part of the code caused at least three times the same bug in the past few months. And encouraged us to start refactoring.

src/lib/BarStore.js
gregorylegarec:
```
return this.fetchComingSoonApps()
  .then(comingSoonApps => comingSoonApp.filter(comingSoonApp => apps.filter(app => app.slug === comingSoonApp.slug))
  .then(filteredComingSoonApps => {
    // Should be handled in a .then() after fetchAppsList call
    // 😭
    this.appsList = apps.concat(comingSoonApps)
    return this.appsList
  })
```

Wow such elegance
.                                                                        very Promise

src/index.jsx
enguerran: I didn't know.

src/index.jsx
enguerran: or we can use a webpack loader and a real file.

src/components/ClaudyMenu.jsx
enguerran: @gregorylegarec try to type 4, 8, 15, 16, 23, and 42 if you're lost.

@CPatchane As we use variables to express intention in our functions, we use components to express intention in our more complex components. As far as I know, expressiveness is far more important than reusability. That is why I suggested to split the big fat component, to increase expressiveness. But as it is just a matter of expressiveness and clean code, and that is not blocking, we can close the discuss and admit we will be slower to add further functionalities. 😜

enguerran:
I commented code, but none of my comments are blocking suggestions. So take the good ones, forget the bad ones and go on.

gregorylegarec:
Happy friday afternoon @CPatchane ! 🎉

PR #83, opened 2017-07-06
fix: filter coming soon app :bug:

kosssi:


src/lib/config.js
```diff
@@ -104,6 +104,8 @@ async function updateAppsItems (config) {
   }
 
   comingSoonApps = await fetchComingSoonApps()
+    // drop coming soon apps already installed
+    .filter(comingSoonApp => apps.filter(app => app.slug === comingSoonApp.slug).length === 0)
```
gregorylegarec: i think that 
```js
.filter(comingSoonApp => apps.find(app => app.slug === comingSoonApp.slug))
```
is better (and clearer)

kosssi: done

gregorylegarec: Oops, it is actually:
```js
.filter(comingSoonApp => !apps.find(app => app.slug === comingSoonApp.slug))
```

PR #85, opened 2017-07-07
Fix bug when apps fetching error + minor UI improvements

CPatchane:


CPatchane:
[Passer le cozy-bar sous preact](https://trello.com/c/y8ZdQ0d2/504-passer-le-cozy-bar-sous-preact)

src/index.jsx
```diff
@@ -52,8 +52,8 @@ const injectBarInDOM = (data) => {
     if (visible) {
       barNode.dataset.drawerVisible = true
     } else {
-      // total drawer closing transition duration = .5s
-      setTimeout(() => { barNode.dataset.drawerVisible = false }, 500)
+      // total drawer closing transition duration = .3s
+      setTimeout(() => { barNode.dataset.drawerVisible = false }, 300)
```
gregorylegarec: Not a big fan on relying on a timeout related to a CSS duration, both values may evolve differently. Instead I suggest to use a TransitionListener. But it will be for a next PR. :)

CPatchane: I agree about that, it would be better handled using listener 👍

CPatchane: @gregorylegarec I pushed a fix for that, thanks 👍

gregorylegarec:
Thanks @CPatchane ! 🌴☀️

PR #87, opened 2017-07-12
fix: Do not open a new tab when switching to another app :wrench:

nono:


gregorylegarec:
Fortunately you're here !

PR #90, opened 2017-07-19
Display Claudy as an intent

CPatchane:


CPatchane:
[(5) [NUA-3] ETU mes applications n'ont pas besoin de permissions pour afficher Claudy](https://trello.com/c/gHTic2pp/501-5-nua-3-etu-mes-applications-nont-pas-besoin-de-permissions-pour-afficher-claudy)

src/lib/intents.js
```diff
@@ -0,0 +1,139 @@
...
+
+      if (handshaken && event.data.type === `intent-${intent._id}:done`) {
+        return resolve(event.data.document)
+      }
+
+      if (!handshaken) {
+        return reject(new Error('Unexpected handshake message from intent service'))
+      }
+
+      // We may be in a state where the messageHandler is still attached to then
```
lespacedunmatin: s/then/the/

src/lib/intents.js
```diff
@@ -0,0 +1,139 @@
...
+      // window, but will not be needed anymore. For example, the service failed
+      // before adding the `unload` listener, so no `intent:cancel` message has
+      // never been sent.
```
lespacedunmatin: s/never/ever/

src/lib/intents.js
```diff
@@ -0,0 +1,139 @@
+import { cozyFetchJSON } from './stack'
+
+/*********************************************
+This following code should never be changed here since it's must be the exact same than in cozy-client-js/src/intents.js
+Service creating functions have been removed since it's not used by the bar
+**********************************************/
```
m4dz: Why not importing / use the cozy-client-js/src/intents.js here rather than duplicate them?

CPatchane: The bar uses an home-made client to interact with the stack, since the `cozy-client-js/src/intents.js` [depends on the client initialised by `cozy-client-js`](https://github.com/cozy/cozy-client-js/blob/master/src/intents.js#L109) I cannot just import it and use it. Do you see another way/solution to not duplicating it?

m4dz: I'll take a look at it, let's keep it as it yet. Maybe we can split tje `intents` lib into commonizable parts, and so have a _cozy-client-js_ flavour, and a vanilla one?

CPatchane: Agree, that could work, thanks 👍

PR #91, opened 2017-07-21
fix: Do not hijack click outside the nav for react apps :fire:

nono:


CPatchane:
Shoud fix the issue, thanks @nono 👍

PR #96, opened 2017-09-12
feat: launcher

y-lohse:
⚠️ DO NOT MERGE ⚠️ 
I need to merge other things before this one, but it's ready for review.

So, the launcher. It's [react-autosuggest](https://github.com/moroshko/react-autosuggest) inside the cozy bar, with suggestions provided over intents.

When the user types something in the search field, apps that have registered themselves are queried for suggestions. Instead of constantly opening and closing intents, they are started at the beginning and kept open the whole time. This is a massive hack of the intent system and should be replaced soon ([full discussion here](https://github.com/cozy/cozy-client-js/issues/201), now is the time to weight in).

I have a PR on the way for Drive that implements a first suggestion source, I'll link to it when it's ready.


src/components/Launcher.jsx
```diff
@@ -0,0 +1,188 @@
...
+    window.addEventListener('message', this.onMessageFromSource)
+    
+    rawIntent(INTENT_VERB, INTENT_DOCTYPE)
+    .then(intent => {
+      const { services } = intent.attributes
+      if (!services) return
+      
+      services.forEach(service => {
+        // we're injecting the iframes, but they will stay invisible
+        const iframe = document.createElement('iframe')
```
goldoraf: I don't like messing with the DOM too much in components: would you mind to instantiate these iframes using JSX in the render method?

src/components/Launcher.jsx
```diff
@@ -0,0 +1,188 @@
...
+    }
+    else {
+      console.log('unhandled message:', event)
+    }
+  }
+  
+  onChange = (event, { newValue }) => {
+    this.setState({
+      query: newValue
+    })
```
enguerran: Could we take the habit to use the functional form of `setState`:

```js
this.setState((state) => ({...state, query: newValue})
```

So `onChange` could be written:

```js
onChange = (event, { newValue: query }) => {
  this.setState(state => ({ ...state, query })
}
```

src/components/Launcher.jsx
```diff
@@ -0,0 +1,188 @@
...
+  onSuggestionsFetchRequested = async ({ value }) => {
+    const promises = []
+    
+    this.sources.filter(source => source.ready).forEach(source => {
+      const promise = new Promise(resolve => {
+        source.resolve = resolve
+        source.window.postMessage({ query: value }, source.origin)
+      })
+      promises.push(promise)
+    })
```
enguerran: This is my favorite. Instead of creating an empty array then a `forEach` that `push` something, why don't we use a beautiful declarative `map` function:

```js
const promises = this.sources.filter(source => source.ready).map(source => 
  new Promise((resolve) => {
    source.resolve = resolve
    source.window.postMessage({ query: value }, source.origin)
  })
)
``` 

Is that more explicit to express the **what** instead of the **how**?

enguerran: Recommended read: https://tylermcginnis.com/imperative-vs-declarative-programming/

src/components/Launcher.jsx
```diff
@@ -0,0 +1,188 @@
...
+    
+    const suggestionsFromSources = await Promise.all(promises)
+    
+    const suggestionSections = suggestionsFromSources.map(response => ({
+      title: this.sources.find(source => source.id === response.id).slug,
+      suggestions: response.suggestions
+    }))
```
enguerran: If you resolve promises just after the promise creation, why don't you use the `then()` chain?

```js
this.sources.filter(source => source.ready).forEach(source =>
  new Promise(resolve => {
    source.resolve = resolve
    source.window.postMessage({ query: value }, source.origin)
  })
  .then(({ id, suggestions }) => ({
    title: this.sources.find(source => source.id === id).slug,
    suggestions
  })
  .then((suggestions) => {
    this.setState((state) => ({ ...state, suggestions }))
  }
)
```

Or something approaching.

src/components/Launcher.jsx
gregorylegarec: I agree with @goldoraf, shouldn't we put this code part into `/lib/intents.js` ? The intent lib in cozy-client actually deals with all the DOM interactions.

src/lib/intents.js
```diff
@@ -1,5 +1,21 @@
 import { cozyFetchJSON } from './stack'
 
+
+// This is a function that does the bare minimum in order to bypass the normal intent flow. To be replaced in th next version of intents.
+export function rawIntent (action, type, data = {}, permissions = []) {
```
gregorylegarec: What the purpose of the rawIntent ? Is it what we called redirection in [this issue](https://github.com/cozy/cozy-client-js/issues/201) ?

src/components/Launcher.jsx
```diff
@@ -0,0 +1,188 @@
...
+    
+    this.setState({
+      suggestions: suggestionSections
+    })
+  }
+  
+  onSuggestionsClearRequested = () => {
+    this.setState({
+      suggestions: []
+    })
```
enguerran: Same old, same old:

```js
onSuggestionsClearRequested = () => {
  this.setState((state) => ({
    ...state,
    suggestions: []
  }))
}
```

src/components/Launcher.jsx
y-lohse: Is there an upside to using the functional form when the new state is *not* computed from the previous state?

src/components/Bar.jsx
```diff
@@ -85,6 +86,9 @@ class Bar extends Component {
         </h1>
         <hr class='coz-sep-flex' />
         {__TARGET__ !== 'mobile' && !isPublic &&
+          <Launcher />
+        }
+        {__TARGET__ !== 'mobile' && !isPublic &&
```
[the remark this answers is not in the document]
y-lohse: I'll double check but I think adding a wrapper messes with the flexbox layout here.
.... which could be fixed, but then again I don't mind the repetition too much here as the launcher block and the navigation block are two very different things.

src/components/Launcher.jsx
y-lohse: I find the async / await form much more readable so I tend to favor it over the `.then` version.

y-lohse:
@enguerran "Launcher" came up during our conversations, but I admit I haven't given the name too much thought. As you point out, I don't think there's an established name for this. Maybe `Searchbox` or `Omnibox`?

src/components/Launcher.jsx
enguerran: The very important upside is `setState` is asynchronous and multiple sequential calls could mislead to unwanted state.
And the corollary is it uses pure function which are easier to maintain.

Even if you don't mutate the state, even if you are quite sure you are not calling `setState` multiple times, you may be homogeneous with the rest of your code and to always use the same syntax/format.

That is why I recommend using the functional `setState` instead of the shorthand version.

src/components/Launcher.jsx
enguerran: If you want to, but my suggest here is not about `then()` against `async/await`.
It's more about the usage of `Promise.all` which await for every single promises to be resolved before continuing.
See https://jsfiddle.net/jjnqLjdo/

src/components/Launcher.jsx
y-lohse: 👍 I'll go with @goldoraf 's suggestion and put it in the render method. I really don't want to legitimize this intent workaround, or encourage anyone else to use it.

src/components/Launcher.jsx
y-lohse: I understand the upside, but in this case the upside doesn't apply? Sequential calls would be perfectly fine here, we always want the latest version of `query` as provided by the AutoSuggest component.

And in terms of homogeneity, I think our codebase favors the shorthand version. I also find it more readable.

src/components/Launcher.jsx
y-lohse: Ah, sorry, I misunderstood. Yes, that makes a lot of sense! I think I meant to do that at some point but it slipped my mind.

src/components/Launcher.jsx
```diff
@@ -0,0 +1,188 @@
...
+          id: intent._id,
+          ready: false,
+          window: null,           // will hold a reference to the window we're sending messages to
+          resolve: null           // a reference to a function to call when the source sends suggestions
+        })
+      })
+    })
+  }
+  
+  onMessageFromSource = (event) => {
```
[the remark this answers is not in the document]
y-lohse: It does multiple things, depending on the type of message, so it's not really possible to express the code's intention in the function name. At best it would be `handleIntentMessage` or something equally not-very-informative.
In this case the `onSomeEvent` form is also in line with the rest of the component (`onSuggestionsFetchRequested` and all from React AutoSuggest).

src/components/Bar.jsx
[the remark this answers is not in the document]
y-lohse: As I thought, the flexbox layout in the bar here doesn't allow to add a wrapper without needing to reorganize the css :/

src/lib/intents.js
y-lohse: Wow, I almost missed this one, sorry! It's actually to accomplish what we called complex communication. So if @CPatchane wants something similar for Claudy, here's an ugly way to do it!
…though we'd better update the spec and the API, really.

y-lohse:
@everyone : thanks for the feedback! I've changed a bunch of things in 5e48af60ea7adc8666551e8d6d954644725a4219 , if you want to have a look.

y-lohse:
The PR I'm waiting for is https://github.com/cozy/cozy-drive/pull/394 since there's no point in having a search bar with no suggestion providers :)

src/components/SearchBar.jsx
```diff
@@ -0,0 +1,187 @@
...
+        
+        this.sources.push({
+          slug: service.slug,     // can be used to show where a suggestion comes from
+          origin: serviceOrigin,
+          id: intent._id,
+          ready: false,
+          window: null,           // will hold a reference to the window we're sending messages to
+          resolve: null           // a reference to a function to call when the source sends suggestions
+        })
+      })
```
enguerran:
```js
this.sources = services.map(service => {
  const url = service.href
  this.setState((state) => ({...state, sourceURLs: [...state.sourceURLs, url]})
  const serviceOrigin = url.split('/', 3).join('/')
  return {
    slug: service.slug,
    origin: serviceOrigin,
    id: intent._id,
    ready: false,
    window: null,
    resolve: null
  }
})
```

That let you remove instantiation `this.sources = []` from the constructor.

That let you remove the constructor in favor of a class properties for setting `this.state`.

src/components/SearchBar.jsx
```diff
@@ -0,0 +1,187 @@
...
+      
+      window.addEventListener('message', this.onMessageFromSource(this.sources))
+    })
+  }
+  
+  onMessageFromSource = (sources) => (event) => {
+    // this re-implements a subset of injectService found in lib/intents, though only the part that are useful for suggestions
+    const source = sources.find(source => source.origin === event.origin)
+    
+    if (!source) return
```
enguerran: `return null`

src/components/SearchBar.jsx
```diff
@@ -0,0 +1,187 @@
...
+          }
+        ]
+      }))
+    })
+  }
+  
+  onSuggestionsClearRequested = () => {
+    this.setState({
+      suggestions: []
+    })
```
enguerran: There is code duplication here:

```js
clearSuggestions = () => {
  this.setState((state) => ({...state, suggestions: []}))
}

// then

onSuggestionsClearRequested = () => {
  this.clearSuggestions()
}

// and

onSuggestionsFetchRequested = ({ value }) => {
  this.clearSuggestions()

  // ...
}
```

src/components/SearchBar.jsx
```diff
@@ -0,0 +1,187 @@
...
+      window.location.href = url
+    }
+    else {
+      console.log('suggestion onSelect (' + onSelect + ') could not be executed')
+    }
+    
+    this.setState({ query: '' })
+  }
+  
+  getSectionSuggestions = (section) => section.suggestions.slice(0, 10) // arbitrary limit, should probably depend on the number of sections
```
enguerran: Don't use unnamed magic number, or in other words, name your magic number to explicit your intention:

```js
const ARBITRARY_LIMIT_DEPENDS_ON_SECTIONS_NUMBER = 10
getSectionSuggestions = (section) => section.suggestions.slice(0, ARBITRARY_LIMIT_DEPENDS_ON_SECTIONS_NUMBER)
```

And magic: there is no need of comments anymore.
```

y-lohse:
@enguerran thanks, very good points, I'll to another pass. 
Though I just wanted to point out that I find it's a good habit to list / initiate the list of class properties in the constructor or somewhere near there, as opposed to adding them "in the middle of nowhere" :)

PR #102, opened 2017-10-03
Email modal

CPatchane:


CPatchane:
[(5) [BAR-18] Bouton d'aide : envoi d'un mail](https://trello.com/c/FHb89K5i/617-5-bar-18-bouton-daide-envoi-dun-mail)

src/components/Nav.jsx
```diff
@@ -33,14 +36,14 @@ class Nav extends Component {
...
         ) {
-        // if it's not a cozy-bar nav popup, close the opened popup
-        if (!this.rootRef.contains(event.target)) {
-          this.setState({ // reset all
-            apps: {busy: false, opened: false},
-            settings: {busy: false, opened: false}
-          })
-        }
-        event.stopPropagation()
+      // if it's not a cozy-bar nav popup, close the opened popup
```
gregorylegarec: Cannot this part be in the Modal component ? Like "any modal close any other opened modal" ?

src/lib/BarStore.js
```diff
@@ -25,6 +25,10 @@ export default class BarStore {
     return createIntent(null, 'CLAUDY', 'io.cozy.settings', data)
   }
 
+  getSupportIntent (data) {
+    return createIntent(null, 'SUPPORT', 'io.cozy.settings', data)
```
gregorylegarec: Just a remark: it feels weird and I am still pretty sure that it is not the way we should use intents. Poke @y-lohse

src/components/Nav.jsx
CPatchane: I think not because the common state is handled in this component. If I put in the modal component, it won't know that another modal is opened, I think.

src/lib/BarStore.js
CPatchane: What seems weird to you? I don't see another way to do it right now

gregorylegarec: That's the problem, we have no other way to do this, so let's keep it. But asking for a `SUPPORT` action on a `io.cozy.settings` doctype (with no id, I guess) really means nothing to me.

src/components/Nav.jsx
gregorylegarec: Maybe with a `static` property ? What thinks @goldoraf about it ?

goldoraf: Well, I'm not a big fan of `static` properties, even if I used then first in the Notifier. I recently tried another approach related to modal management, using redux of course: https://github.com/cozy/cozy-drive/blob/master/src/lib/react-cozy-helpers/ModalManager.jsx

The idea is to attach the ModalManager in your layout (https://github.com/cozy/cozy-drive/blob/master/src/drive/components/Layout.jsx#L22) and to dispatch a `showModal` action like this: https://github.com/cozy/cozy-drive/blob/master/src/drive/ducks/files/Container.jsx#L89

You could easily enhance the ModalManager so that it could handle multiple modals ;)

src/lib/BarStore.js
CPatchane: I agree with your point, maybe we could discuss about that later with the team.

src/components/Nav.jsx
CPatchane: That seems a good idea indeed, but for an application only. Since the bar don't and won't use redux I would prefer to keep the current way which seems to be better and simpler.

gregorylegarec:
Thanks @CPatchane ! 🍶

PR #104, opened 2017-10-16
Expose a method to update cozy-bar locale without reloading

CPatchane:


gregorylegarec:
I guess that it is not possible to keep the `lang` as a state property and re-render `index.jsx` when `lang` is updated ?

CPatchane:
Here we go! I finally found a working way to use props rendering, no tricky things anymore 🎉 🎉 🎉

PR #109, opened 2017-10-30
Bar hooks with redux

ptbrowne:
Exposes API to control the bar left/center/right content
JS API :

```js
const { setBarLeft } = cozy.bar

setBarLeft("<span>Hello</span>")
```

Also `setBarRight` and `setBarCenter`

React API :

```js
const { BarLeft } = cozy.bar
const MyComponent = <div>
  <BarLeft>Hello</BarLeft>
</div>
```

First PR with without redux : https://github.com/cozy/cozy-bar/pull/108

### DIFF
--- a/src/components/Bar.jsx
+++ b/src/components/Bar.jsx
@@ -2,6 +2,8 @@
 /* global __PIWIK_TRACKER_URL__  __PIWIK_SITEID__ __PIWIK_DIMENSION_ID_APP__ */
 
 import React, { Component } from 'react'
+import { connect } from 'react-redux'
+
 import { translate } from '../lib/I18n'
 import { shouldEnableTracking, getTracker, configureTracker } from 'cozy-ui/react/helpers/tracker'
 
@@ -10,6 +12,7 @@ import Nav from 'components/Nav'
 import SearchBar from 'components/SearchBar'
 import Claudy from 'components/Claudy'
 import SupportModal from 'components/SupportModal'
+import { getContent } from 'lib/reducers'
 
 class Bar extends Component {
   constructor (props, context) {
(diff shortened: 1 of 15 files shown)


src/components/Bar.jsx
```diff
@@ -117,19 +121,29 @@ class Bar extends Component {
...
 }
 
-export default translate()(Bar)
+const mapStateToProps = state => ({
+  barLeft: getContent(state, 'left'),
+  barRight: getContent(state, 'right'),
+  barCenter: getContent(state, 'center')
+})
+
+export default compose(
```
CPatchane: Why using `compose` (and its dependency) here? Can't we do:
```js
expose default translate()(connect(mapStateToProps, null)(Bar))
```

ptbrowne: I find the `compose` version more readable.

ptbrowne: I do not think there is a problem with having `lodash` as a dependency.

CPatchane: No problem with `compose` from `lodash`, but since we don't use it anywhere else for now (and we can do it without), I would prefer without, thanks 👍

CPatchane:
Thanks @ptbrowne 👍

src/components/Bar.jsx
gregorylegarec: I agree with @CPatchane, I am not a big fan of dependencies, especially lodash. I think we should not add a dependency for just one "comfort function".

However, if we really want to use `compose` for exporting our components, we should maintain code coherence and use it everywhere. And in every app.

ptbrowne: It is not really adding a dependency since you have already 32 dependencies that depends on it.

Why are you not a fan of lodash ?

gregorylegarec: I personally think that it's just a lib of shortcuts hiding intersting JS behavior and patterns.

But I am not against using it. But if we do so, it should be done in all our components in every app.

PR #110, opened 2017-10-30
feat: toggle searchbar with hidden code

y-lohse:
The searchbar is still hidden by default, but can be forced to appear by placing a magic string in the URL's search part.


CPatchane:
Thanks @y-lohse 🎉

PR #111, opened 2017-11-02
Fix bar

ptbrowne:
- [x] visibility of the bar was taken from the store instead of state
- [x] drawer was not shown if BarRight was used
- [x] transifex was not used in travis


CPatchane:
Thanks @ptbrowne 🎉

PR #112, opened 2017-11-02
chore: define process.env.NODE_ENV to production

ptbrowne:
Removing warning "You are currently using minified code outside of NODE_ENV === 'production'. This means that you are running a slower development build of Redux."


CPatchane:
Thanks @ptbrowne 👍

PR #113, opened 2017-11-03
fix: use babel-preset-cozy-app instead of using babel-polyfill

ptbrowne:
babel-polyfill was already imported in `cozy-client-js` and it would lead to a bug in Safari :

`undefined is not an object` when trying to call toString on the Int8Array.

Removing `babel-polyfill` completely could lead to problems so we turn to `babel-preset-env` in `babel-preset-cozy-app` to polyfill only what we need.


CPatchane:
Thanks @ptbrowne 🎉

PR #118, opened 2017-11-08
Reduce item size in apps lists :lipstick:

CPatchane:
Before
screen shot 2017-11-08 at 11 13 31
After
screen shot 2017-11-08 at 11 18 34


CPatchane:
[(2) [BAR-12] Diminuer la taille des icones dans le menu mobile](https://trello.com/c/GF0fVq0W/661-2-bar-12-diminuer-la-taille-des-icones-dans-le-menu-mobile)

PR #120, opened 2017-11-09
feat: highlight search matches in results

y-lohse:


src/components/SearchBar.jsx
```diff
@@ -7,13 +7,68 @@ const INTENT_VERB = 'OPEN'
...
+    })
+
+    currentIndex += resultTerm.length + spacerLength
+  })
+
+  // matchedIntervals can overlap, so we merge them.
+  // - sort the intervals by starting index
+  // - add the first interval to the stack
+  // - for every interval,
+  // - - add it to the stck if it doesn't overlap with the stack top
```
CPatchane: s/stck/stack ?

y-lohse: 👍

src/components/SearchBar.jsx
```diff
@@ -7,13 +7,68 @@ const INTENT_VERB = 'OPEN'
...
+  const normalizedQueryTerms = normalizeString(query)
+  const normalizedResultTerms = normalizeString(searchResult)
+
+  const matchedIntervals = []
+  const spacerLength = 1
+  let currentIndex = 0
+
+  normalizedResultTerms.forEach(resultTerm => {
+    normalizedQueryTerms.forEach(queryTerm => {
+      const index = resultTerm.indexOf(queryTerm)
```
enguerran: I would prefer a `regex` search

src/components/SearchBar.jsx
```diff
@@ -7,13 +7,68 @@ const INTENT_VERB = 'OPEN'
...
+      if (index >= 0) {
+        matchedIntervals.push({
```
enguerran: don'tlike `forEach` with side effects. Can't we have a `map`?

src/components/SearchBar.jsx
```diff
@@ -7,13 +7,68 @@ const INTENT_VERB = 'OPEN'
...
+    currentIndex += resultTerm.length + spacerLength
+  })
+
+  // matchedIntervals can overlap, so we merge them.
+  // - sort the intervals by starting index
+  // - add the first interval to the stack
+  // - for every interval,
+  // - - add it to the stck if it doesn't overlap with the stack top
+  // - - or extend the stack top if the start overlaps and the new interval's top is bigger
+  const mergedIntervals = matchedIntervals.sort((intervalA, intervalB) => intervalA.from > intervalB.from).reduce((computedIntervals, newInterval) => {
```
enguerran: I guess the algorithme is not mastered as it needs some comments.

src/components/SearchBar.jsx
y-lohse: That's subjective though. I prefer `indexOf` for searches that don't involve pattern matching.

src/components/SearchBar.jsx
y-lohse: No, precisely because of the "side effects", since they aren't side effects at all. They aren't a byproduct of the `forEach`, they are the reason for it.
More extensively, I'm creating new information based on 2 existing arrays, not reducing or transforming the existing data. The array I'm building could have less, the same, or more elements than any of the base arrays. And the data it contains is not a transformation of the values of the existing arrays either.

src/components/SearchBar.jsx
enguerran: did you benchmark it, I guess regex are way faster than indexOf: https://jsperf.com/regexp-indexof-perf

y-lohse: From what I've seen, it depends on the browser and on the test case.
In the test case you've linked, only `test` is used, whereas here we want the index of the match, so `String.search` would be more appropriate. It's also doing the tests on node class names, which we're really not doing here. [This is another benchmark](https://jsperf.com/exec-vs-match-vs-test-vs-search/5) that I just ran, and `indexOf` just blows everything off the water (849,050,703 ops vs 17,738,919 for `search`and `20,586,521` for `test`).

src/components/SearchBar.jsx
y-lohse: I found the algorithm not so trivial, enough to warrant a quick explanation in plain english.

src/components/SearchBar.jsx
enguerran: ok

src/components/SearchBar.jsx
enguerran: When you say:
> No, precisely because of the "side effects", since they aren't side effects at all. 

You mean the code could be refactored like this:
```js
normalizedResultTerms.forEach(newFunction1())

function newFunction1() {
  return resultTerm => {
    normalizedQueryTerms.forEach(newFunction(resultTerm));
    currentIndex += resultTerm.length + spacerLength;
  };
}

function newFunction(resultTerm) {
  return queryTerm => {
    const index = resultTerm.indexOf(queryTerm)
    if (index >= 0) {
      matchedIntervals.push({
        from: currentIndex + index,
        to: currentIndex + index + queryTerm.length
      })
    }
  }
}
```
with `newFunction` and newFunction1` as [pure functions](https://en.wikipedia.org/wiki/Pure_function) ?

I mean: you need to be opinionated to write some code. But the reason you give to justify the point is not good, nor real.
Nevertheless you can keep this code instead of writing a real pure functional version. There is no problem to have this code on the codebase. It is just a matter of readability and maintainability.

PR #121, opened 2017-11-10
Fix broken claudy

CPatchane:


src/components/Bar.jsx
```diff
@@ -101,8 +101,8 @@ class Bar extends Component {
   }
 
   renderRight () {
-    const { usageTracker, claudyOpened,
-      enableClaudy, fireClaudy, displayOnMobile, isPublic } = this.props
+    const { displayOnMobile, isPublic } = this.props
+    const { enableClaudy, fireClaudy, claudyOpened, usageTracker } = this.state
```
ptbrowne: Sorry for the mixup between `state` and `props` :/.

I think what we can improve from this is the naming of the properties. When I read `enableClaudy`, I read as a verb + noun so I think it is a function, so I guess it is coming from the props. I think for boolean value like `displayOnMobile` the name should be `displayedOnMobile`.

CPatchane: No worries, I agree these parts can be a bit confusing. We could be it's not the priority for now^^

CPatchane:
[Claudy n'est plus affiché](https://trello.com/c/k9es7shr/682-claudy-nest-plus-affich%C3%A9)

PR #126, opened 2017-11-20
fix: hide apps button on mobile

y-lohse:
The bar has an option called `displayOnMobile` which is false by default. It used to hide pretty much everything, [including the button to show the other apps](https://github.com/cozy/cozy-bar/blob/a03cd1ab40fb09b8c0fd5a9552e3f77961c650e8/src/components/Bar.jsx#L95-L97).

This changed after dc3a3657fa17d98447eb3339912c946d684da50b , and now this button is always displayed (which is a problem in Drive). I would like to revert to the previous behavior.

That being said, I wonder if this change would cause problems for Cozy Bank?

### DIFF
--- a/src/components/Bar.jsx
+++ b/src/components/Bar.jsx
@@ -93,11 +93,11 @@ class Bar extends Component {
   }
 
   renderLeft () {
-    const { t } = this.props
+    const { t, displayOnMobile, isPublic } = this.props
     // data-tutorial attribute allows to be targeted in an application tutorial
-    return <button className='coz-bar-burger' onClick={this.toggleDrawer} data-icon='icon-apps' data-tutorial='apps-mobile'>
+    return (__TARGET__ !== 'mobile' || displayOnMobile) && !isPublic ? <button className='coz-bar-burger' onClick={this.toggleDrawer} data-icon='icon-apps' data-tutorial='apps-mobile'>
       <span className='coz-bar-hidden'>{t('drawer')}</span>
-    </button>
+    </button> : null
   }
 
   renderRight () {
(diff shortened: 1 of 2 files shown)


y-lohse:
[🐛📱 L'icône du menu plateforme est affiché dans l'app](https://trello.com/c/8jxcX1ai/585-%F0%9F%90%9B%F0%9F%93%B1-lic%C3%B4ne-du-menu-plateforme-est-affich%C3%A9-dans-lapp)

ptbrowne:
We use `displayOnMobile: true` so we should be safe if we revert to the previous behavior.

That being said, I think we should go for a `displayOnMobile: true` by default since Drive behavior will not be the norm in Cozy apps. For those kind of API changes, I think we should go first for a "deprecated feature" log and then change it some time later (2 months) should be fine.

src/components/Bar.jsx
```diff
@@ -93,11 +93,11 @@ class Bar extends Component {
   }
 
   renderLeft () {
-    const { t } = this.props
+    const { t, displayOnMobile, isPublic } = this.props
     // data-tutorial attribute allows to be targeted in an application tutorial
-    return <button className='coz-bar-burger' onClick={this.toggleDrawer} data-icon='icon-apps' data-tutorial='apps-mobile'>
+    return (__TARGET__ !== 'mobile' || displayOnMobile) && !isPublic ? <button className='coz-bar-burger' onClick={this.toggleDrawer} data-icon='icon-apps' data-tutorial='apps-mobile'>
```
ptbrowne: Why is `public` mentioned ?

y-lohse: This is the exact condition that was previously used. IIRC, we also hide this button if we're on a public page, ie. the public page for a shared album.

y-lohse:
Ok perfect, I think that makes sense. I'll add the deprecated message.

PR #133, opened 2017-11-22
Fix bar elements

ptbrowne:
Before

image

After

image

### DIFF
--- a/src/styles/base.css
+++ b/src/styles/base.css
@@ -4,7 +4,6 @@
   flex: 1 0;
 }
 
-
 [role=banner] [data-icon] {
   background-repeat: no-repeat;
   background-position: 0 50%;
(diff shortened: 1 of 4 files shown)


src/styles/base.css
```diff
@@ -42,3 +41,8 @@
 [role=banner] [data-icon='icon-claudy'] {
   background-image: url(../assets/icons/16/icon-claudy.svg)
```
CPatchane: Missing `;`

ptbrowne: ? `;` are missing everywhere in this file. It does not matter, it is on the last line.

ptbrowne: Anyway, I added them in https://github.com/cozy/cozy-bar/pull/133/commits/a22bfd1d8d721a9d4470db35afc3ffcc6a4ae8b9

y-lohse:
Please ignore, I just need a url for this image.

cozy-bar-triplet

PR #134, opened 2017-11-22
feat: add icon to searchbar result

enguerran:
Related to https://github.com/cozy/cozy-drive/pull/509

### DIFF
--- a/src/components/SearchBar.jsx
+++ b/src/components/SearchBar.jsx
@@ -193,6 +193,7 @@ class SearchBar extends Component {
 
   renderSuggestion = (suggestion) => (
     <div className='coz-searchbar-autosuggest-suggestion-content'>
+      {suggestion.icon && <img className='coz-searchbar-autosuggest-suggestion-icon' src={suggestion.icon} alt='icon' />}
       <div className='coz-searchbar-autosuggest-suggestion-title'>
         {highlightQueryTerms(suggestion.title, this.state.query)}
       </div>
--- a/src/styles/searchbar.css
+++ b/src/styles/searchbar.css
@@ -95,6 +95,7 @@
   border-bottom-width: 1px;
   border-style:  solid;
   border-color: var(--silver);
+  cursor: pointer;
 }
 [role=banner] .coz-searchbar-autosuggest-suggestion:last-child {
   border-bottom-width: 0;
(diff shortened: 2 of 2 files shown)


enguerran:
[(2) Ajouter les icônes dans les résultats de la recherche](https://trello.com/c/QYT8Opun/562-2-ajouter-les-ic%C3%B4nes-dans-les-r%C3%A9sultats-de-la-recherche)

src/components/SearchBar.jsx
```diff
@@ -193,6 +193,7 @@ class SearchBar extends Component {
 
   renderSuggestion = (suggestion) => (
     <div className='coz-searchbar-autosuggest-suggestion-content'>
+      {suggestion.icon && <img src={suggestion.icon} />}
```
y-lohse: Can you add an `alt` attribute? Maybe something like "suggestion icon"? It's not great but better than nothing.

enguerran:
@GoOz est-ce que tu vois quelque chose que je pourrais ajouter pour l'intégration de cette image dans la liste des résultats ?

src/components/SearchBar.jsx
GoOz: Or just `alt=""`. Better nothing than a useless alternative text.

GoOz:
@enguerran Un peu de margin-right sur l'image, genre 1rem je dirais, à vue d'œil. Mais c'est tout.

src/components/SearchBar.jsx
y-lohse: > Better nothing than a useless alternative text.

Really though? "suggestion icon" wouldn't be entirely useless, even though it would be the same for every suggestion.
Going further — these icons aren't really content, so maybe they shouldn't be in the DOM in the first place? They might be better as background images, no?

GoOz: Nope, no background, ``is fine, better even. 
In that case, I would say that the best description to the image is the name of the file (because we don't have title, complete description or anything else about this image but the filename) and there's no need to repeat the filename in the `alt` because it's right after.
The `alt` attribute is not supposed to tell you what kind of image this is or what is its purpose but what's inside, what it represents. And we're not gonna do that. So I would say that the best value for the `alt` is no value at all.
Actually when in doubt, you should let the `alt` empty.

y-lohse: Meh. Yes, `alt` should tell you what the image represents, and in this case it's "an icon for the file's mime type". So the ideal `alt` would be something like "icon of an image".
But ok, we don't have that info, so I get the leaving it blank. 

I'm still not sure `img` is better than a background image. The icons are visual enhancers, they are hardly content. But they do have a meaning so they aren't entirely visual either. Anyway I'm cool with an empty `alt`.

enguerran:
@y-lohse I added some `margin-right`

PR #152, opened 2017-12-18
Add index tests + fix locale change by connecting I18n

CPatchane:


CPatchane:
[[Cozy-barre] Ne prend pas en compte le changement de langue](https://trello.com/c/6lt6Erki/742-cozy-barre-ne-prend-pas-en-compte-le-changement-de-langue)

src/lib/BarStore.js
```diff
@@ -191,7 +193,10 @@ export class Provider extends Component {
     this.store = props.store
   }
 
-  render ({children}) {
+  render () {
+    const { children } = this.props
```
ptbrowne: why don't you keep it in the arguments ?

src/lib/BarStore.js
```diff
@@ -191,7 +193,10 @@ export class Provider extends Component {
...
     return (children && children[0]) || null
   }
 }
+
+Provider.childContextTypes = { barStore: PropTypes.object }
```
ptbrowne: can you add isRequired ?

src/index.jsx
```diff
@@ -60,18 +60,28 @@ const injectBarInDOM = (data) => {
...
+  const EnhancedI18n = connect(state => ({
+    lang: getLocale(state)
+  }))(I18n)
+
+  const barComponent = (
     <BarProvider store={barStore}>
       <ReduxProvider store={reduxStore}>
-        <I18n
+        <EnhancedI18n
           lang={getLocale(reduxStore.getState())}
```
ptbrowne: If I am not mistaken, the lang can be removed here

test/index.spec.js
```diff
@@ -0,0 +1,27 @@
...
+      lang: 'en',
+      iconPath: '',
+      cozyURL: 'https://mock.cozy',
+      token: 'mock',
+      displayOnMobile: true
+    }
+    const Bar = init(options)
+    const mounted = mount(Bar)
+    expect(mounted.getElement()).toMatchSnapshot()
+  })
```
ptbrowne: Can you add a test when you change the language ?

src/lib/BarStore.js
CPatchane: That's a preact way, not a React way so it would break all tests that use React.

ptbrowne: Ah yes I see.

test/index.spec.js
CPatchane: I didn't find a good way to test that case here for now. The mounted bar here doesn't give me a full rendered bar as we could expect. So there is no way to get the difference between a lang an another.

test/__snapshots__/index.spec.js.snap
```diff
@@ -0,0 +1,42 @@
...
+        "replaceReducer": [Function],
+        "subscribe": [Function],
+        Symbol(Symbol.observable): [Function],
+      }
+    }
+  >
+    <Connect(I18n)
+      dictRequire={[Function]}
+    >
+      <Unknown
```
ptbrowne: Unknown ?

CPatchane: I know it's not great :/, but that tests is just to check the bar first root rendering (it was broken by some previous PRs). When I will have more time I will try to improve them, but I am opened for some improvement ideas :)

test/index.spec.js
ptbrowne: What is missing in the rendered component here is only the lang ? It seems in the test that the store has `lang: undefined`

CPatchane: I was trying to test directly the rendered parts or components. I don't wanted to test directly the store since it will repeat [that store test](https://github.com/cozy/cozy-bar/blob/master/test/lib/api.spec.js#L47) I think.

ptbrowne: ok

test/__snapshots__/index.spec.js.snap
ptbrowne: I think the time is now, you will never have more time.

gregorylegarec:
Thanks @CPatchane 👍

PR #153, opened 2017-12-18
Fix support modal broken

CPatchane:


CPatchane:
[Modales Cassées](https://trello.com/c/BY1BaN7j/739-modales-cass%C3%A9es)

ptbrowne:
I personally prefer when there is a description to a PR, it is easier to understand.

src/styles/supportModal.css
```diff
@@ -2,6 +2,10 @@
   min-height: 8em;
 }
 
+[role=banner] .coz-support-modal {
+  max-width: inherit;
```
ptbrowne: inherit ? Why not none here ?

ptbrowne: Also I do not understand why everywhere in bar you use [role=banner] and not something like "cozy-bar".

CPatchane: You're right, none is a better value 👍 
For the [role="banner"] it's an aria value for the role attribute (https://www.w3.org/TR/wai-aria/#landmark_roles)

ptbrowne: Yeah I know it's aria stuff but it still feels weird to me to do this instead of `.cozy-bar`.

ptbrowne:
LGTM

src/styles/supportModal.css
CPatchane: Since we also use [role=application] for the application part, it's consistent for me.

PR #158, opened 2018-01-04
feat: add onLogOut option on init

drazik:
In bank, we need to do some custom things when we log out of the mobile app. So we needed to pass an `onLogOut` handler to the cozy-bar. That's the purpose of this PR.

### DIFF
--- a/src/components/Bar.jsx
+++ b/src/components/Bar.jsx
@@ -103,7 +103,7 @@ class Bar extends Component {
   renderRight = () => {
     const { displayOnMobile, isPublic, renewToken } = this.props
     return (__TARGET__ !== 'mobile' || displayOnMobile) && !isPublic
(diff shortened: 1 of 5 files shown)


src/components/Settings.jsx
```diff
@@ -60,7 +60,7 @@ const Settings = ({ t, onLogOut, settingsData, onClaudy, isDrawer = false, isCla
     </ul>
     <ul className='coz-nav-group'>
       <li className='coz-nav-item'>
-        <button role='menuitem' data-icon='icon-logout' onClick={onLogOut} title={t('logout')}>
+        <button role='menuitem' data-icon='icon-logout' onClick={onLogOut} type='button' title={t('logout')}>
```
kosssi: pourquoi ce changement ?

drazik: Résidu de debug, je vire ça :)

PR #175, opened 2018-02-27
chore: release v4.8.9 :fire:

y-lohse:


CPatchane:
@y-lohse could you also add `Remove -moz-scrollbars-none usage for firefox` to the changelog please before releasing it? Thanks 👍

y-lohse:
@CPatchane yeah sure but can you tell me what the actual impact of that is ? Are we showing the scrollbars on firefox again ? Or was it just not needed ?

CPatchane:
We wanted to hide the scroll bar, but the previous way disabled scrolling behaviour on Firefox so we reverted the part related to Firefox.

PR #180, opened 2018-03-09
feat: debounce search results

enguerran:
Put a debounce in the sent message to get search results.
But it appears there is a lot of obscur code in there and in cozy-drive intent.

### DIFF
--- a/src/components/SearchBar.jsx
+++ b/src/components/SearchBar.jsx
@@ -1,6 +1,7 @@
 import React, { Component } from 'react'
 import { translate } from 'cozy-ui/react/I18n'
 import Autosuggest from 'react-autosuggest'
+import debounce from 'lodash.debounce'
 import { fetchRawIntent } from '../lib/intents'
 
 const INTENT_VERB = 'OPEN'
@@ -95,7 +96,8 @@ class SearchBar extends Component {
           }
         })
 
-        window.addEventListener('message', this.onMessageFromSource(this.sources))
+        const debounced = debounce(this.onMessageFromSource(this.sources), 250, { 'maxWait': 1000 })
+        window.addEventListener('message', debounced)
       })
   }
 
--- a/package.json
+++ b/package.json
@@ -86,6 +86,7 @@
     "core-js": "^2.5.3",
     "cozy-client-js": "^0.3.19",
     "localforage": "^1.5.5",
+    "lodash.debounce": "4.0.8",
     "node-polyglot": "^2.2.2",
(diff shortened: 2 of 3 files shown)


enguerran:
Is this working? I tried to comment this line with no change in behavior
https://github.com/cozy/cozy-drive/blob/911f9c6865313d436a70cb038a99b9f0285615ef/src/drive/ducks/services/components/SuggestionProvider.jsx#L19

enguerran:
I found there is a `fetch` query here
https://github.com/cozy/cozy-drive/blob/911f9c6865313d436a70cb038a99b9f0285615ef/src/drive/ducks/services/components/SuggestionProvider.jsx#L49-L52
but when I comment this line, there is another request to `data/io.cozy.files`, I don't know where it comes from.

enguerran:
This component https://github.com/cozy/cozy-drive/blob/911f9c6865313d436a70cb038a99b9f0285615ef/src/drive/ducks/services/components/SuggestionProvider.jsx#L7 has no render function. What is its purpose?

enguerran:
I don't think this code is a very good one to handle asynchronous calls https://github.com/cozy/cozy-bar/blob/df2763c8e8f659ad87d2711864d04c70af759cb6/src/components/SearchBar.jsx#L114-L118

enguerran:
I don't know why travis is ko https://travis-ci.org/cozy/cozy-bar/builds/351399941#L794

CPatchane:
It's ko because of bundlesize, I limited the bundled size of the bar to 80kb and it begins to be higher than this limit.

src/components/SearchBar.jsx
```diff
@@ -95,7 +96,8 @@ class SearchBar extends Component {
           }
         })
 
-        window.addEventListener('message', this.onMessageFromSource(this.sources))
+        var debounced = debounce(this.onMessageFromSource(this.sources), 250, { 'maxWait': 1000 })
```
y-lohse: Shouldn't we use `const` instead of `var` here?

enguerran: han, pris en flag de copy/paste development ! Déso, je change ça, évidemment qu'on reste sur du `const`.

enguerran:
[Lors de la première requête de recherche, les résultats sont incorrects](https://trello.com/c/ZAiViH5Y/853-lors-de-la-premi%C3%A8re-requ%C3%AAte-de-recherche-les-r%C3%A9sultats-sont-incorrects)

PR #184, opened 2018-03-13
fix: debounce suggestion fetch requests

y-lohse:
In #180 , we debounced the function that receives the search results. This worked to a degree, but not quite, because sometimes the results come in one after the other with a delay greater than 250ms, and we still get the flashes of results.

Instead, I suggest we debounce the calls to the function that makes the query to the services. So if the user quickly types "hey what's up", we only fetch the suggestions once, when he/she's done typing.

It's still possible to get flashes of course, if you wait just over 250ms and start typing again. But it's less frequent.

### DIFF
--- a/src/components/SearchBar.jsx
+++ b/src/components/SearchBar.jsx
@@ -96,9 +96,10 @@ class SearchBar extends Component {
           }
         })
 
-        const debounced = debounce(this.onMessageFromSource(this.sources), 250, { 'maxWait': 1000 })
-        window.addEventListener('message', debounced)
+        window.addEventListener('message', this.onMessageFromSource(this.sources))
       })
+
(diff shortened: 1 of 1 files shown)


y-lohse:
[Lors de la première requête de recherche, les résultats sont incorrects](https://trello.com/c/ZAiViH5Y/853-lors-de-la-premi%C3%A8re-requ%C3%AAte-de-recherche-les-r%C3%A9sultats-sont-incorrects)

enguerran:
Okay, my bad.
J'ai cherché sur l'API de AutoSuggest voir si y'avait pas un truc pour debouncer les input de l'utilisateur.
Bref, ça me semble une bonne remarque et une excellente PR.

PR #205, opened 2018-05-22
Improve CGU banner

CPatchane:


CPatchane:
[(0,5) Bandeau CGUs au dessus de la cozy-barre](https://trello.com/c/5owwZkfn/991-05-bandeau-cgus-au-dessus-de-la-cozy-barre)

src/components/Banner.jsx
```diff
@@ -9,7 +9,9 @@ class Banner extends Component {
   }
 
   componentDidMount () {
-    this.setState(() => ({ unmounted: false }))
+    return setTimeout(() => {
```
gregorylegarec: ⚠️SETTIMEOUT ALERT ⚠️
What it is used for ?

gregorylegarec: Ok, so this is used to animate the bar. As it is not very self explicit, I suggest using something like:

```jsx
animate() {
  // To animate we have to use a setTimeout to force a CSS class update and trigger CSS animation
  return setTimeout(() => this.setState({ unmounted: false }))
}

componentDidMount () {
  this.animate()
}
```

PR #207, opened 2018-05-24
Temporary get back editor as name prefix

CPatchane:


CPatchane:
It's temporary because we have no way to deal without a name prefix with that.

PR #213, opened 2018-06-25
chore: update UI to v10.0.1

GoOz:
Update UI so the bar's drawer always goes over the Nav on mobile


GoOz:
[Drawer from cozy-bar goes under Nav](https://trello.com/c/SKucZoFe/182-drawer-from-cozy-bar-goes-under-nav)

Issue #216, opened 2018-07-31
Can't disconnect on mobile browsers


GoOz:
The drawer menu needs a safe space so the user can click on the last item of the menu, the disconnect button in our case, especially on iPhone X but any mobile when minimal UI is triggered too.

On minimal UI, the Disconnect button is in the safe area therefore it's not clickable:

<img width="377" alt="screenshot 2018-08-06 14 20 29" src="https://user-images.githubusercontent.com/1280069/43716331-2068ea84-9984-11e8-8296-ba3472dcbb5a.png">

And when minimal UI is not triggered, the button is hidden underneath the Safari UI therefore it's not clickable either.

<img width="388" alt="screenshot 2018-08-06 14 20 07" src="https://user-images.githubusercontent.com/1280069/43716334-2368bd68-9984-11e8-9525-2772b5166860.png">

ptbrowne:
> "Can't disconnect on minimal UI is trigger and on iPhone X"

I don't understand the title of the issue.

I don't understand the problem ? Is the drawer not scrollable ?

PR #218, opened 2018-08-06
fix(Drawer): Fix drawer position on mobile devices

GoOz:
Also make nav elements in drawer clickable while adding safe space on Iphone X and prevent scrolling background on iOS
Fixes: #216

⚠️ Side effect: Displaying the drawer will switch off minimal UI. There's no way I can bypass that but it seems like a small price to pay.

### DIFF
--- a/src/styles/drawer.css
+++ b/src/styles/drawer.css
@@ -1,9 +1,9 @@
 [role=banner] .coz-drawer-wrapper {
-  position: absolute;
+  position: fixed;
   top: 0;
   left: 0;
   width: 100vw;
-  height: 100vh;
+  height: 100%;
   display: flex;
   visibility: visible; /* overwrite default [aria-hidden=true] style */
 }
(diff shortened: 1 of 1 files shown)


src/styles/drawer.css
```diff
@@ -135,3 +135,12 @@
   padding: 2em 2em .5em;
   margin: 0;
 }
+
+[role=banner] .coz-drawer--settings {
+  padding-bottom: env(safe-area-inset-bottom);
+}
+
+/* /!\ Trick to prevent application from scrolling in the background when the drawer is opened */
+[role=banner][data-drawer-visible=true] + [role=application] {position: fixed;
```
ptbrowne: Add carriage return before `position: fixed`

GoOz:
Is it OK for you guys @ptbrowne  @CPatchane  ?

src/styles/drawer.css
```diff
@@ -135,3 +135,13 @@
   padding: 2em 2em .5em;
   margin: 0;
 }
+
+[role=banner] .coz-drawer--settings {
+  padding-bottom: env(safe-area-inset-bottom);
```
CPatchane: TIL

src/styles/drawer.css
```diff
@@ -135,3 +135,13 @@
...
+}
+
+/* /!\ Trick to prevent application from scrolling in the background when the drawer is opened */
+[role=banner][data-drawer-visible=true] + [role=application] {
+  position: fixed;
+  width: 100%;
```
CPatchane: Is this styles not breaking applications layout?

GoOz: Not as far as I tested, no. 
The thing is that the drawer cover 90% of the application so you don't have much to see of it.

CPatchane: Let's try then

PR #219, opened 2018-08-07
Improvements on app list

ptbrowne:
The goal of this PR is to have the app list available in offline mode.
The apps already being saved in local storage thanks to redux persist,
the only need was to remove the spinner from being displayed if the apps
had already been fetched once. I took this opportunity to add tests
on the app reducer and to simplify it a bit.

The redux-persist functionality was a bit "hidden" by the fact that it was
only on the mobile side, I think there is no problem to have it also in browser,
this is why I removed the branching into two different stores.


src/lib/reducers/apps.js
```diff
@@ -11,19 +11,28 @@ const EXCLUDES = ['settings', 'onboarding']
...
+const onMobile = __TARGET__ !== 'mobile'
+export const getApps = (state, mobile = onMobile) => {
+  if (!state.apps) return []
+
+  if (!mobile) {
+    return state.apps
   }
 
-  if (!state.apps || !state.apps.data) return []
-  return state.apps.data.filter(app =>
```
CPatchane: Is normal thate here you don't use the data property anymore? I don't see it anywhere else

ptbrowne: It is normal the apps.data was not useful and was complicating the code. The `apps` from the action are directly stored under the `apps` property of this slice.

ptbrowne:
I have removed `localForage` and used directly `localStorage`. This way, the bundle is no longer > 100kb.

PR #220, opened 2018-08-08
refactor: AppIcon

ptbrowne:
In preparation for another PR, I've refactored the `AppIcon` code.

* `AppIcon` understands directly the `app` object

The AppList had 6 props related to the `app` object. The `AppList`
had to know all this properties, and forced the creation of useless
props on the `AppIcon`.

[Components/functions should know as little as possible of each other
API][Demeter].

Now the `AppList` [only responsibility][responsibility] is simply to render a list of `AppIcons`.
The `AppIcon` responsibility is to render the `Icon` and text directly from
the `app` object.

* Threading of the `toggleComingSoon`

The `toggleComingSoon` prop was passed from component to component since
the state of the modal was stored in the `Bar` state. It forced every component
in the middle to know about this function that they did not care about.

Putting the `ComingSoonModal` inside the `AppIcon` allows to remove this
unnecessary threading. To bypass the fact that the Modal would be rendered
inside the `Drawer` window, I added `into='body'` which renders the modal
into the `Body`.

* Added tests

[Demeter]:  https://fr.wikipedia.org/wiki/Loi_de_D%C3%A9m%C3%A9ter
[responsibility]: https://fr.wikipedia.org/wiki/Principe_de_responsabilit%C3%A9_unique

### DIFF
--- a/src/components/AppIcon.jsx
+++ b/src/components/AppIcon.jsx
@@ -37,7 +37,7 @@ class AppIcon extends React.Component {
     const label = (app.namePrefix ? (app.namePrefix + ' ') : '') + app.name
     const comingSoon = app.comingSoon || false
     const iconSrc = app.icon && app.icon.cached ? app.icon.src : defaultIcon
-    const canShowComingSoonDescription = HAS_COMING_SOON_DESCRIPTION[app.slug]
+    const canShowComingSoonDescription = HAS_COMING_SOON_DESCRIPTION.hasOwnProperty(app.slug)
 
     let appClass = comingSoon ? 'coz-bar-coming-soon-app' : ''
     let onClick = canShowComingSoonDescription ? this.openComingSoon : null
--- a/src/components/AppIcon.jsx
+++ b/src/components/AppIcon.jsx
@@ -2,39 +2,43 @@ import React from 'react'
 import { translate } from 'cozy-ui/react/I18n'
 import defaultIcon from '../assets/icons/16/icon-cube-16.svg'
 
-export const AppIcon = ({ t, onClick, app }) => {
-  const dataIcon = app.icon ? `icon-${app.slug}` : ''
-  const blurry = !app.icon || !app.icon.cached || false
-  const label = (app.namePrefix ? (app.namePrefix + ' ') : '') + app.name
-  const comingSoon = app.comingSoon || false
-  const iconSrc = app.icon && app.icon.cached ? app.icon.src : defaultIcon
+class AppIcon extends React.Component {
+  render () {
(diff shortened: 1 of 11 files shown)


test/jestLib/setup.js
```diff
@@ -1,7 +1,12 @@
...
+
+import 'babel-polyfill'
 
 // polyfill for requestAnimationFrame
 /* istanbul ignore next */
 global.requestAnimationFrame = (cb) => {
   setTimeout(cb, 0)
 }
+
+Enzyme.configure({ adapter: new Adapter() })
```
CPatchane: oh I didn't know configuring Enzyme inside jest setup would just work as simplely, thanks for this tip! ❤️

src/components/AppIcon.jsx
```diff
@@ -1,30 +1,81 @@
...
-  )
+
+  render () {
+    const { t, app } = this.props
+    const dataIcon = app.icon ? `icon-${app.slug}` : ''
+    const blurry = !app.icon || !app.icon.cached || false
+    const label = (app.namePrefix ? (app.namePrefix + ' ') : '') + app.name
+    const comingSoon = app.comingSoon || false
+    const iconSrc = app.icon && app.icon.cached ? app.icon.src : defaultIcon
+    const canShowComingSoonDescription = HAS_COMING_SOON_DESCRIPTION[app.slug]
```
CPatchane: I would use `hasOwnProperty` here to avoid getting true with a property from prototype but it's clearly not blocking

src/components/AppIcon.jsx
```diff
@@ -1,30 +1,81 @@
...
-  )
+
+  render () {
+    const { t, app } = this.props
+    const dataIcon = app.icon ? `icon-${app.slug}` : ''
+    const blurry = !app.icon || !app.icon.cached || false
+    const label = (app.namePrefix ? (app.namePrefix + ' ') : '') + app.name
+    const comingSoon = app.comingSoon || false
+    const iconSrc = app.icon && app.icon.cached ? app.icon.src : defaultIcon
+    const canShowComingSoonDescription = HAS_COMING_SOON_DESCRIPTION.hasOwnProperty(app.slug)
```
drazik: Is there a reason why you store the coming soon apps in an object instead of an array, since you're just checking the properties of the object ?

ptbrowne: I prefer to check for inclusion in an HashMap than a list.

* Inclusion in a list is O(n)
* Inclusion in a hashmap is O(1)

At this size though, the performance is negligible.

PR #228, opened 2018-08-28
Build before shasum

gregorylegarec:
Just tired of having bad shasum in my pull requests


CPatchane:
Thanks @gregorylegarec ☯️

PR #230, opened 2018-09-12
Feat: new inter-apps navigation (+ some cleaning)

CPatchane:
Cleaning:
- Remove FakeAppsList with forbidden exception usgae
- Remove coming soon apps
- Better files outline

__BREAKING CHANGES__: 
- A completely new way to navigate through applications handling the current application and the Home application, the current app slug is now necessary and must be passed from the `cozy-stack` to the application to use this new navigation
- The `editor` isn't used as the app name prefix anymore. If you want a name prefix to be displayed int the bar, please use the `namePrefix` field in the app manifest


gregorylegarec:
> A completely new way to navigate through applications handling the current application and the Home application

It's no really a breaking change, is it ?

src/lib/reducers/apps.js
```diff
@@ -11,20 +11,27 @@ const EXCLUDES = ['settings', 'onboarding']
...
+  app.slug === state.appSlug
 
 // selectors
 const onMobile = __TARGET__ !== 'mobile'
 export const getApps = (state, mobile = onMobile) => {
   if (!state.apps) return []
 
+  const appsWithCurrentApp = state.apps.map(app => {
+    if (isCurrentApp(state, app)) {
+      return Object.assign({}, app, { isCurrentApp: true })
```
gregorylegarec: Is there a reason here to not use spread operator ?

src/lib/reducers/apps.js
```diff
@@ -104,9 +102,16 @@ const reducer = (state = defaultState, action) => {
     case FETCH_APPS:
       return { ...state, isFetching: true }
     case RECEIVE_APP_LIST:
-      return { ...state, isFetching: false, hasFetched: true, apps: action.apps }
-    case RECEIVE_APP_LIST_FORBIDDEN:
-      return { ...state, isFetching: false }
+      const appsList = action.apps.map(app => {
+        if (isCurrentApp(state, app)) {
+          return Object.assign({}, app, { isCurrentApp: true })
```
gregorylegarec: Spread operator ?

src/locales/en.json
```diff
@@ -13,7 +13,9 @@
 
   "menu": {
     "apps": "Apps",
-    "settings": "Settings"
+    "settings": "Settings",
+    "home_mobile": "Back to home...",
```
gregorylegarec: Are `...` necessary ? It feels weird reading it here. Maybe in the UI it feels better.

src/components/Apps/AppNavButtons.jsx
```diff
@@ -0,0 +1,45 @@
...
+import { connect } from 'react-redux'
+
+import { getHomeApp } from 'lib/reducers'
+
+import Icon from 'cozy-ui/react/Icon'
+import homeIcon from 'assets/icons/icon-cozy-home.svg'
+
+class AppNavButton extends Component {
+  render () {
+    const { homeApp, busy, onClick, appName, appNamePrefix, iconPath, opened } = this.props
```
gregorylegarec: Global remark:

I don't feel realy comfortable having a ambiguous var name here, as `home` means the _Home page_, but also it the slug of our `Home app`. So it feels we're using directly a reference to the Home app, as we did in the past with Collect for example, and it's a code smell.

But I also agree that `home` is also the best word to describe this variable. What alternative do we have ? `default` ? `entry` ? `main` ?

src/index.jsx
```diff
@@ -117,7 +117,7 @@ const getDefaultIcon = () => {
 
 const getAppNamePrefix = () => {
   const appNode = document.querySelector(APP_SELECTOR)
-  return appNode.dataset.cozyAppNamePrefix || appNode.dataset.cozyAppEditor || undefined
+  return appNode.dataset.cozyAppNamePrefix || undefined
```
gregorylegarec: It's a bad pattern to return `undefined`, as `undefined` means that a variable did not get value assigned. Instead, we should return `null` or an empty string `''`.

gregorylegarec:
It looks great @CPatchane

src/locales/en.json
CPatchane: I follow the mockup for this text, I think we will discuss again about that wording

CPatchane:
I forgot to precise that the new navigation need that a new attribute `data-cozy-app-slug` must be used by applications to provided the current application slug from the `cozy-stack` to that application. Without that, the application won't be broken but we won't have the wanted navigation UX.
Also, this big update is a good reason to bump a major version and stop using the `editor` as name prefix, and that change is clearly a breaking change.

PR #231, opened 2018-09-14
Upgrade webpack to v4 + add webpack-analyzer usage

CPatchane:
Use webpack 4 to get tree-shaking on lodash usage (previously all included in the build)
Add an anlyzer when providing `ANALYZER=true` when running build script


gregorylegarec:
Great job 💯

PR #263, opened 2018-09-26
Last fixes

CPatchane:
* Apps was mapped after a process (set home application)
* Hover style was not correct on mobile drawer
* Don't replace home name
* Force home icon

### DIFF
--- a/src/components/Apps/AppItem.jsx
+++ b/src/components/Apps/AppItem.jsx
@@ -55,7 +55,7 @@ export class AppItem extends React.Component {
   }
 
   render () {
-    const { app } = this.props
+    const { app, icon } = this.props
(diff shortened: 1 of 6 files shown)


src/components/Apps/AppItem.jsx
```diff
@@ -78,12 +78,21 @@ export class AppItem extends React.Component {
...
-            className='coz-nav-apps-item-icon'
-            fetchIcon={stack.get.icon}
-            key={app.slug}
-          />
+          {icon
+            ? <img
+              src={icon}
+              className='coz-nav-apps-item-icon'
+              alt=''
+              width='32'
```
gregorylegarec: The dimensions are handler by the CSS

PR #273, opened 2018-09-27
Fix home btn hover style issue

CPatchane:
Remove blue left border for the home button on hover


CPatchane:
Alright, do you prefer this way @gregorylegarec? :P

PR #284, opened 2018-10-11
Fix locales handling

CPatchane:
SET_LANG is the action to be dispatched by Settings when language changing


CPatchane:
[(𝐗𝐋) ETU je peux naviguer facilement dans mes applications](https://trello.com/c/iYsSBZio/275-%F0%9D%90%97%F0%9D%90%8B-etu-je-peux-naviguer-facilement-dans-mes-applications)

PR #322, opened 2018-11-01
feat: Display current app on mobile

kosssi:
After talk with @Claiw on the last meeting, the current app should be display on mobile app like web.


ptbrowne:
👍

PR #339, opened 2018-11-12
Handled global React as aliases instead

CPatchane:
Better handling React or Preact in the bar and its submodules
No need to use as global variable in the code anymore thanks to aliases


gregorylegarec:
Great, thanks !

PR #341, opened 2018-11-13
Use yarn alias instead of require node_modules

nono:


CPatchane:
TIL npm aliases, thanks a lot @nono! 🌮 
I just pushed some changes to fix the tests with these changes.

PR #342, opened 2018-11-13
chore: Remove preact dep

ptbrowne:
If I am not mistaken, preact is no longer a dependency as the bar
bundles React by default.

React also should not be a dependency as the bar is bundled. It should be a dev dependency ?


CPatchane:
Thanks @ptbrowne 👍

PR #344, opened 2018-11-14
Handle modules only on init by the app

CPatchane:
By this way, whatever you charge the bar before or after the application script, the bar will look for React/ReactDOM in global scope only on init() for its working (and will start serving the API only after the init).
Should make the bar more robust for its usage.

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -2,10 +2,6 @@
 
 'use strict'
 
-import React from 'react'
-import { render } from 'react-dom'
-
-import I18n from 'cozy-ui/react/I18n'
 import stack from './lib/stack'
 import {
   deleteApp,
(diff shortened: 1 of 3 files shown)


src/index.jsx
```diff
@@ -197,17 +206,46 @@ const init = async ({
...
   })
 }
 
 const updateAccessToken = accessToken => {
   stack.updateAccessToken(accessToken)
 }
 
+// Handle exceptions for API before init
+const apiReferences = {}
+APILocations.forEach(location => {
```
CPatchane: We use references here, it allows to get the method available when the application will 'import/require' it but it will look for the real `exposedAPI` only on demand when the application will call the method.

CPatchane:
@y-lohse Completely agree! I'll write something about all of this 👍

CPatchane:
@gregorylegarec inject via `init()` can't be possible since we have to replace also in the cozy-bar modules (react-redux for example)

PR #346, opened 2018-11-14
perf: Do not render nav part when not in mobile

ptbrowne:
The apps where rendered two times. In the drawer part and in the nav part.

### DIFF
--- a/src/components/Bar.jsx
+++ b/src/components/Bar.jsx
@@ -6,6 +6,8 @@ import { connect } from 'react-redux'
 
 import { translate } from 'cozy-ui/react/I18n'
 import Icon from 'cozy-ui/react/Icon'
+import withBreakpoints from 'cozy-ui/react/helpers/withBreakpoints'
+
 import {
   shouldEnableTracking,
   getTracker,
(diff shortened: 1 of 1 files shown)


CPatchane:
This change make the `replaceTitleOnMobile` parameter useless. We have to see if this parameter is still working and if we still need it. If not we will be able to remove it completely and do this change.

ptbrowne:
> This change make the replaceTitleOnMobile parameter useless. We have to see if this parameter is still working and if we still need it. If not we will be able to remove it completely and do this change.

I don't understand, what can I do to have this PR merged ?

ptbrowne:
Ping ?

CPatchane:
Didn't get time to check if this PR doesn't add regressions or to check if `replaceTitleOnMobile` is still by some apps or not, sorry :/

Crash--:
In Drive, we have a few `replaceTitleOnMobile`. I don't know exactly what this props does?

nono:
Ping @ptbrowne @Crash--

Crash--:
Just took a look this morning and it seems that we do not render (I mean, I don't find it in the DOM) this nav on mobile web, right? Even if I have replaceTitle true/false. But it seems that we use BarRight / Left / Center. 



So I don't konw if it's needed anymore. But if yes, I think we can merge it without breaking anything on drive / contact 🤞

nono:
Ping @ptbrowne

ptbrowne:
I am leaving this open as I think it could be important.

CPatchane (commented):
🤷

PR #352, opened 2018-11-19
Improve history handling for api store

CPatchane:
Use a list instead of a single value to handle API previous contents

### DIFF
--- a/src/lib/api/index.js
+++ b/src/lib/api/index.js
@@ -1,7 +1,10 @@
 import React, { Component } from 'react'
-import { getContent, setContent, setLocale } from '../reducers'
+import { setContent, unsetContent, setLocale } from '../reducers'
 
 import { locations, getJsApiName, getReactApiName } from 'lib/api/helpers'
+
+// The React API need unique IDs, so we will increment this variable
+let idToIncrement = 0
 
 /**
  * Wraps argument into a React element if it is a string. Is used
@@ -27,7 +30,7 @@ const wrapInElement = v => {
 const barContentComponent = (store, location) =>
   class BarContent extends Component {
     componentDidMount() {
-      this.prev = getContent(store.getState(), location)
+      this.componentId = idToIncrement++
       this.setContent(this.props.children)
     }
 
@@ -36,11 +39,15 @@ const barContentComponent = (store, location) =>
         content = React.Children.only(content)
         // eslint-disable-next-line no-empty
       } catch (e) {}
-      store.dispatch(setContent(location, content))
+      store.dispatch(setContent(location, content, this.componentId))
+    }
+
+    unsetContent() {
+      store.dispatch(unsetContent(location, this.componentId))
     }
 
     componentWillUnmount() {
-      this.setContent(this.prev)
+      this.unsetContent()
     }
 
     componentDidUpdate(prevProps) {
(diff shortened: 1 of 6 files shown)


ptbrowne:
Can you add a test :) ?

CPatchane:
[[BAR/STO] Le redimensionnement casse le contenu de la bar sur le store](https://trello.com/c/4COpuzYo/494-bar-sto-le-redimensionnement-casse-le-contenu-de-la-bar-sur-le-store)

CPatchane:
Oops, thanks for the reminder 👍

src/lib/reducers/content.js
```diff
@@ -15,15 +15,25 @@ const defaultState = {
...
+    return { ...state, [action.location]: newState }
   } else {
     return state
   }
 }
 
 // selectors
-export const getContent = (state, location) => state[location]
+export const getContent = (state, location) =>
+  Array.isArray(state[location]) ? state[location].pop() : null
```
ptbrowne: C'est surprenant qu'une méthode appelée `get` ait un side effect.

src/lib/reducers/content.js
```diff
@@ -15,15 +15,25 @@ const defaultState = {
...
-    action.type === SET_CONTENT &&
-    state[action.location] !== action.content
-  ) {
-    return { ...state, [action.location]: action.content }
+  if (action.type === SET_CONTENT) {
+    if (!action.location) return state
+    const newState = Array.isArray(state[action.location])
+      ? state[action.location].slice()
+      : []
+    // avoid doublon
```
ptbrowne: Pourquoi avoid les doublons ?

CPatchane: This part was already in this API, i judge it necessary to still be handled, do you think this part is useless here?

src/lib/reducers/content.js
CPatchane: It will be better with a `unsetContent` handled by the reducer itself, and this function will be without side effect

src/lib/reducers/content.js
ptbrowne: Mmh I don't see why its there, I think it was more to fix the shortcomings of the `this.prev` method. I do not see why duplicates should be ignored.

```
<BarLeft>A</BarLeft>
<BarLeft>A</BarLeft>
<BarLeft>B</BarLeft>
```

If you unmount the 2nd and 3rd <BarLeft />, you will have the Bar displaying nothing.

CPatchane:
Now the store will use a `Map()` object to handle all the content correctly. The content can be replaced on component updates and unset on component onmount.
I also added more automated things in tests :)

src/lib/reducers/content.js
CPatchane: Now store handle correctly components update with content replacing in the Map, no need this check anymore

src/lib/api/index.js
```diff
@@ -1,8 +1,11 @@
 import React, { Component } from 'react'
-import { getContent, setContent, setLocale } from '../reducers'
+import { setContent, unsetContent, setLocale } from '../reducers'
 
 import { locations, getJsApiName, getReactApiName } from 'lib/api/helpers'
 
+// The React API need unique IDs, so we use an iterator here
+let IDiteror = 0
```
ptbrowne: It is not an iterator (nor an iteror :D).

I would just call this `let id = 0`

CPatchane: I would prefer `idToIncrement` to be more explicit, but agree about the iterator naming here. Thanks!

PR #359, opened 2018-11-26
Add apps reducer to redux-persist config

CPatchane:
Allow offline usage with the apps list


CPatchane:
[[BAR] Offline sur les mobiles](https://trello.com/c/5cBYZHfG/491-bar-offline-sur-les-mobiles)

PR #363, opened 2018-11-28
fix: Banks on iOS has not the same appId

Crash--:
Banks has not the same appId on iOS and and Android. This PR fix this issue


CPatchane:
Thanks @Crash-- 👍

PR #374, opened 2018-12-05
Minor fixes

CPatchane:
* Do not display the app name prefix if this is the platform home application
* Use the correct blue color on visited home button
* Disable home button on public pages

### DIFF
--- a/src/components/Apps/AppNavButtons.jsx
+++ b/src/components/Apps/AppNavButtons.jsx
@@ -48,7 +48,10 @@ class AppNavButton extends Component {
 
     return (
       <div className={`coz-nav-apps-btns${isHomeApp ? ' --currentHome' : ''}`}>
-        <a href={homeApp && homeApp.href} className="coz-nav-apps-btns-home">
+        <a
+          href={!isPublic && homeApp && homeApp.href}
+          className="coz-nav-apps-btns-home"
+        >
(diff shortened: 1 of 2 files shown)


src/components/Apps/AppNavButtons.jsx
```diff
@@ -48,7 +48,10 @@ class AppNavButton extends Component {
 
     return (
       <div className={`coz-nav-apps-btns${isHomeApp ? ' --currentHome' : ''}`}>
-        <a href={homeApp && homeApp.href} className="coz-nav-apps-btns-home">
+        <a
```
gregorylegarec: Even with an empty `href` attribute, this `<a />` will still be clickable.

I suggest instead:
```jsx
const homeHref = !isPublic && homeApp && homeApp.href
// ...
{ homeHref ?
   <div href={homeHref} className="coz-nav-apps-btns-home">
      <HomeIcon currentColor />
   </div>
   : <span className="cozy-nav-apps-btns-home">
       <HomeIcon currentColor />
     </span>
}
```
We repeat a little code but I think using a more advanced pattern does not worth it here.

CPatchane: Just remove the `href` here is enough since you can detect the miss of the attribute for css and for the browser an `<a>` without href is "kind like" just a text at the end

PR #388, opened 2018-12-07
Home application fix

CPatchane:
* Disable correctly the link on public page
* Persist context to avoid getting a persisted home app without context and home information


gregorylegarec:
Thanks @CPatchane 👌

PR #403, opened 2018-12-16
feat: Bar should use transpiled version of ui

Crash--:
Since UI is now shipped in a transpiled way we have two choices : 
- use the transpiled version of UI in the bar 
- use the non transpiled version of UI in the bar, but the bar should transpile it 

I just changed the icons since this is the only place I have an error and I want this fix to be merged ASAP, but we should consider to move to the first solution ?


Crash--:
Since I use the transpiled version and the non transpiled one, I have added some ko to the bundle... 

Should I change all the import to `cozy-ui/transpiled/react`?

nono:
Ping @CPatchane

PR #406, opened 2018-12-17
Correctly determine cozy-bar URL usage

CPatchane:
Without the `//` the URL doesn't work on browsers

### DIFF
--- a/src/lib/stack.js
+++ b/src/lib/stack.js
@@ -190,18 +190,16 @@ const determineURL = (cozyURL, ssl) => {
   let url
   let host
   const protocol = ssl ? 'https' : 'http'
+
   try {
+    // only on mobile we get the full URL with the protocol
     host = new URL(cozyURL).host
-    if (host) {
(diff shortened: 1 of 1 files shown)


src/lib/stack.js
```diff
@@ -191,16 +191,18 @@ const determineURL = (cozyURL, ssl) => {
   let host
   const protocol = ssl ? 'https' : 'http'
   try {
+    // only on mobile we get the full URL with the protocol
     host = new URL(cozyURL).host
     if (host) {
       url = `${protocol}://${host}`
     } else {
-      url = cozyURL
+      // on web we just get the domain without protocol
```
gregorylegarec: This part is confusing, what do you think about: 
```js
try { 
  host = new URL(cozyURL).host
  url = !!host && `${protocol}://${host}`
} catch (e) {}

host = host || cozyURL
url = url || `//${host}`

return { COZY_URL: url, COZY_HOST: host }
```

PR #409, opened 2018-12-18
Always use the protocl for COZY_URL by default

CPatchane:
URL beginning by `//` was broking realtime, since we got now the SSL information we can always use the protocol to build the COZY_URL

### DIFF
--- a/src/lib/stack.js
+++ b/src/lib/stack.js
@@ -130,31 +130,8 @@ export const getAppIconProps = () => {
       }
 }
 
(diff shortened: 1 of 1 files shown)


CPatchane:
Closed to use the `cozy-urls` lib instead https://github.com/cozy/cozy-libs/pull/191/files

gregorylegarec:
From where the ssl information comes from ? Why `domain` disappeared ?

PR #429, opened 2019-01-28
fix(Modal): Put Modal into a dedicated place insteed of body

sizvix:
Due to React and Preact difference, where React append the destination element with the Modal and Preact replace the content of the destination by the Modal, we change the destination which was body to a dedicated node.

### DIFF
--- a/src/components/Bar.jsx
+++ b/src/components/Bar.jsx
@@ -186,6 +186,7 @@ class Bar extends Component {
     } = this.props
     return (
       <div className="coz-bar-wrapper">
+        <div id="cozy-bar-modal-dom-place" />
         <div className="coz-bar-container">
           {barLeft || this.renderLeft()}
           {barCenter || this.renderCenter()}
--- a/src/components/SupportModal.jsx
+++ b/src/components/SupportModal.jsx
@@ -34,7 +34,7 @@ class SupportModal extends Component {
         <Modal
           secondaryAction={this.props.onClose}
           className="coz-support-modal"
-          into="body"
+          into="#cozy-bar-modal-dom-place"
         >
           <ModalContent className="coz-support-modal-wrapper">
             <div className="coz-support-modal-content">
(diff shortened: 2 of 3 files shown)


sizvix:
[[Home] Je ne peux plus ouvrir le formulaire d'aide](https://trello.com/c/WFE7Gwxo/675-home-je-ne-peux-plus-ouvrir-le-formulaire-daide)

src/components/Bar.jsx
```diff
@@ -186,6 +186,7 @@ class Bar extends Component {
     } = this.props
     return (
       <div className="coz-bar-wrapper">
+        <div id="CozyBarSupportModalPlace" />
```
CPatchane: Since we will have to use it for all cozy-bar modals, maybe we can name it just `CozyBarModalDomPlace` ?

y-lohse:
> and Preact replace the content of the destination by the Modal

Are you sure it does? We portal stuff into `body` a lot and it works fine.

sizvix:
@y-lohse : there is a bug with Preact apps, ( the home or drive for example, when you go to settings > Help ) , put Modal to Body make some bugs ... So it need a place to be layed ...

src/components/Bar.jsx
sizvix: Why not, if we never have more than one modal at the same time ( that should be OK, but ... )

CPatchane: The bar is not designed to displayed multiple modals but it could have different modals ;)

gregorylegarec: Great you found the solution.

I think that the `id` should not be CamelCased but use hypen, like our CSS classes and other `id` attributes in the Cozy-Bar codebase ([example](https://github.com/cozy/cozy-bar/blob/master/src/components/Apps/index.jsx#L67)).

CPatchane:
Thanks @sizvix ! 👍

PR #434, opened 2019-01-29
Use relative node_modules in webpack config

CPatchane:
Webpack process differently absolute path vs relative path for `node_modules`, it's better to avoid absolute path.


gregorylegarec:
So, why are we changing relative path to absolute path here ? And it's not really absolute.

PR #436, opened 2019-02-01
fix: Import transpiled UI

Crash--:
Let's see the bundle size. 

Seems to work on : 
- Drive web (responsive & desktop) 
- Drive mobile 

~~Need to test on : 
Drive mobile~~


CPatchane:
I think it would be better to use a webpack alias for that `cozy-ui/react/` -> `cozy-ui/transpiled/react`. What do you think?

Crash--:
> I think it would be better to use a webpack alias for that cozy-ui/react/ -> cozy-ui/transpiled/react. What do you think?

I'm not confortable with that since the code you read isn't the code you really think you read without checking your webpack conf... But it's not my project. 

I just want my application to run where we want. If you prefer an alias, set an alias, I don't care.

CPatchane:
I'll do a PR for that. Thanks @Crash--.

PR #437, opened 2019-02-01
Use transpiled cozy-ui components

CPatchane:
Use transpiled cozy-ui components

Tested on cozy-store

@Crash-- Please test again on Drive browser and mobile to merge this PR.


Crash--:
Thank you! Seems to work well on Drive

PR #441, opened 2019-02-07
fix: :wrench: Make webpack use the postcss config

nono:


config/webpack.config.extract.js
```diff
@@ -23,7 +23,9 @@ module.exports = ({ filename }) => ({
             loader: 'postcss-loader',
             options: {
               config: {
-                path: path.join(__dirname, '../postcss.config.js')
```
nono: Postcss-load documentation says to not do that (it should be empty, or the path of a directory where the config file can be found if you really have some specific).

postcss.config.js
```diff
@@ -1,23 +1,23 @@
 const path = require('path')
 
-const { production } = require('./config/webpack.vars')
```
nono: It was broken since we moved from `NODE_ENV` to webpack env: `production` was always undefined as we didn't give any env to `webpack.vars`.

postcss.config.js
```diff
@@ -1,23 +1,23 @@
...
+      path: [path.resolve(__dirname, 'src/styles/')]
+    }),
+    require('autoprefixer')(['last 2 versions', 'not dead', 'not ie <= 11'])
+  ]
 
-const commons = [
-  require('postcss-import')({
-    path: [path.resolve(__dirname, 'src/styles/')],
-    plugins: production ? [
-      require('autoprefixer')(['last 2 versions']),
```
nono: It seems that giving plugins to `postcss-import` doesn't work: the autoprefixer was adding the prefixes to the generated CSS.

nono:
[Corriger la config webpack/postcss de la cozy-bar](https://trello.com/c/J3cMAuLa/944-corriger-la-config-webpack-postcss-de-la-cozy-bar)

PR #447, opened 2019-02-08
feat: Add style on mobile device

kosssi:
Je rapatrie le style de la cozy-bar qui est encore dans cozy-ui, voir la suppression dans la PR suivante https://github.com/cozy/cozy-ui/pull/832

### DIFF
--- a/src/styles/bar.css
+++ b/src/styles/bar.css
@@ -10,6 +10,17 @@
   font-size: 1rem;
 }
 
+@media (max-width: 64em) {
+  [role=banner] {
+    position: fixed;
+    top: 0;
+    left: 0;
+    width: 100%;
+    background-color: white;
+    box-sizing: border-box;
+  }
+}
+
 [role=banner] .coz-bar-wrapper {
   display: flex;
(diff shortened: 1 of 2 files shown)


src/styles/bar.css
```diff
@@ -10,6 +10,17 @@
   font-size: 1rem;
 }
 
+@media (max-width: 64em) {
+  [role=banner] {
```
CPatchane: So, no need to have the double selector anymore? Good thing if so 👍

kosssi: I don't understand why we need double selector before, it's ok in banks with this.

CPatchane: Do you try on a native mobile browser? @GoOz seems to say that it was specifically for that case

GoOz: It was clearly a conflict between Cozy-bar styles and UI styles on mobile browsers when minimal-UI was deployed. 
I don't remember clearly what was the bug but it sure be cool to test this on native browser on mobile indeed.

PR #451, opened 2019-02-11
feat: Add setTheme to custum cozy bar design

kosssi:
Change theme bar
---

It's possible to update design on the cozy-bar with `setTheme` function.

```jsx
const { setTheme } = cozy.bar
setTheme('default')
setTheme('primary')
```

This feature impact only mobile version.

default | primary
-------|--------
capture d ecran 2019-02-11 a 16 58 31 | capture d ecran 2019-02-11 a 16 58 12


test/components/__snapshots__/Bar.spec.jsx.snap
```diff
@@ -2,7 +2,7 @@
 
 exports[`Bar should display the Searchbar 1`] = `
 <div
-  className="coz-bar-wrapper"
+  className="coz-bar-wrapper theme-undefined"
```
kosssi: @CPatchane you understand why it's `undefined`?

edas:
Instead of 
```
html {
  --primary: var(--dodgerBlue)
} 
[role=banner] .coz-bar-wrapper {
  background-color: var(--white)
}
[role=banner] .coz-bar-wrapper.theme-primary {
    background-color: var(--primary)
}
```


Shouldn't we have something like : 
```
html {
  --primary: var(--dodgerBlue)
  --background: var(--white)
} 
[role=banner] .coz-bar-wrapper {
  background-color: var(--background)
}
[role=banner] .coz-bar-wrapper.theme-primary {
    --primary: var(--white)
    --background: var(--primary)
}
```

?

src/components/Bar.jsx
```diff
@@ -185,7 +187,7 @@ class Bar extends Component {
       userActionRequired
     } = this.props
     return (
-      <div className="coz-bar-wrapper">
+      <div className={`coz-bar-wrapper theme-${theme}`}>
```
CPatchane: I'd prefer `coz-theme-XX` to explicit something Cozy specific here

src/lib/reducers/index.js
```diff
@@ -1,6 +1,7 @@
 import { combineReducers } from 'redux'
 import * as content from './content'
 import * as locale from './locale'
+import * as design from './design'
```
CPatchane: Wy `design` and not just `theme`? Design term seems very vague to me here

test/components/__snapshots__/Bar.spec.jsx.snap
CPatchane: Because we don't test against connected components but we used the dumb component https://github.com/cozy/cozy-bar/blob/master/test/components/Bar.spec.jsx#L2, you have to add a new props and check props change in the unit tests
Could you also add tests for this new API method? It should the same than [for `setLocale`](https://github.com/cozy/cozy-bar/blob/master/test/lib/api.spec.js#L104-L112)

src/assets/icons/16/icon-apps-primary.svg
```diff
@@ -0,0 +1,5 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 16 16">
```
kosssi: Je pense que c'est problématique d'avoir les SVG avec de la couleur on devrait pouvoir avoir la même SVG pour les 2 themes. Mais c'est à voir dans de l'optimisation dans un futur plus ou moins proche...

CPatchane: Did you try to use the `fill` css property? Or maybe a SVGr component like for [Home icon](https://github.com/cozy/cozy-bar/blob/master/src/components/Apps/IconCozyHome.jsx)?

kosssi: It's more complex because we used it on (background-image](https://github.com/cozy/cozy-bar/pull/451/files#diff-2cb59efd08fdd9e7511322df3541313eR21). 

With my comment I wanted to warn you but not necessarily solve the problem in this PR

CPatchane: Maybe we can avoid using it as a background-image? Did you try it?
This issue should be solved in this PR instead of keeping it in the code for a while IMHO

drazik: @CPatchane I used an `Icon` from cozy-ui instead of the background-image svg. Is it OK for you like this?

PR #455, opened 2019-02-12
Fix support modal styles + upgrade cozy-ui to v19

CPatchane:


CPatchane:
[[Home] Je ne peux plus ouvrir le formulaire d'aide](https://trello.com/c/WFE7Gwxo/675-home-je-ne-peux-plus-ouvrir-le-formulaire-daide)

src/styles/index.styl
```diff
@@ -1,3 +1,4 @@
+@import '../../node_modules/cozy-ui/transpiled/stylesheet.css'
```
Crash--: Should we use ~ instead ?

src/styles/supportModal.css
```diff
@@ -9,7 +9,11 @@
 }
 
 .coz-support-modal .coz-support-modal-wrapper {
-  margin-top: 1.4rem;
+  margin-top: 1rem;
```
Crash--: We have some classes in UI that can help you to remove this one. 

You should use `u-mt-1` and remove this new classe ?

PR #459, opened 2019-02-14
Use timeout drawer closing on app switch

CPatchane:
Better UX with a setimeout :)

### DIFF
--- a/src/components/Apps/AppItem.jsx
+++ b/src/components/Apps/AppItem.jsx
@@ -49,6 +49,10 @@ export class AppItem extends React.Component {
     }
   }
 
+  componentWillUnmount() {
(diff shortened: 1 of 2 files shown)


src/components/Apps/AppItem.jsx
```diff
@@ -58,18 +58,27 @@ export class AppItem extends React.Component {
     }
   }
 
+  onAppSwitch = () => {
+    const { onAppSwitch } = this.props
+    if (typeof onAppSwitch === 'function') {
+      setTimeout(() => {
```
Crash--: You should store this timeout and cleat it on willUnmount(). 

It'll prevent possible memory leaks and action on a non mounted component

gregorylegarec:
Great job CPATCHANE

PR #461, opened 2019-02-15
Feat/inject cozy client

edas:
L'objectif est de permettre d'injecter directement cozy-client dans la barre plutôt que de transmettre des token et recréer un mini client interne


gregorylegarec:
Quick review : totally the right direction 👍

src/lib/stack-client.js
```diff
@@ -0,0 +1,236 @@
...
+  NotFoundException,
+  MethodNotAllowedException,
+  UnavailableStackException,
+  UnavailableSettingsException,
+  UnauthorizedStackException
+} from './exceptions'
+
+import getIcon from './icon'
+
+let cozyClient = undefined
```
drazik: Nit: `let cozyClient` alone initializes the variable to `undefined`

src/lib/stack-client.js
```diff
@@ -0,0 +1,236 @@
...
+    }
+  }
+
+  cozyClient.getStackClient().fetch(method, path, body, options).then(
+    resp => {
+      if (typeof errorStatuses[resp.status] === 'function') {
+        throw new errorStatuses[resp.status]()
+      }    
+      const contentType = resp.headers.get('content-type')
+      const isJson = contentType && contentType.indexOf('json') >= 0
```
drazik: Nit: `contentType.includes('json')` is more readable IMHO (it works everywhere except in IE)

src/lib/stack-client.js
```diff
@@ -0,0 +1,236 @@
...
+      return resp.json()
+    }
+  )
+}
+
+// Cache a request 
+// with a defaultValue in case of 404
+const withCache(fn, defaultValue) {
+  let cache = undefined;
+  const is404 = function(error) {
```
drazik: I think this function can be extracted so we don't create a new function on each `withCache` call

src/lib/stack-client.js
```diff
@@ -0,0 +1,236 @@
...
+const defaultQuota = 10**12 // 1 Tera
+
+// Storage/quota usage
+const getStorageData = function() {
+  return fetchJSON('GET', '/settings/disk-usage').then( 
+    usage => { 
+      return {
+        usage: parseInt(json.data.attributes.used, 10),
+        // TODO Better handling when no quota provided
+        quota: parseInt(json.data.attributes.quota, 10) || defaultQuota,
```
drazik: Maybe just add a `isNaN` check?

src/lib/stack-client.js
edas: ✅ I am requesting more comments. I used to be explicit but if the cozy crowd prefer being implicit, that would be fine with me

src/lib/stack-client.js
edas: ✅ it was taken from the legacy code and not rewritten, I'll do that

src/lib/stack-client.js
edas: ✅ will do (even if in this case the function will be called only one)

src/lib/stack-client.js
edas: It was from the legacy code. I fear they expect the number to be in a string

@gregorylegarec ?

gregorylegarec: We do https://github.com/cozy/cozy-stack/blob/master/docs/settings.md#response

edas:
I welcome any help to see how to test this PR.

My solution is to initialize a new blank app, initialize a cozy client, initialize the bar with it, put my app in a cozy… I am sure one of you could look at it more quickly

src/lib/normalize-url.js
```diff
@@ -0,0 +1,40 @@
+
+// Mobile apps give us full URL with domain and protocol.
+// Web apps could also initilize with a full URL
```
ptbrowne:
```suggestion
// Web apps may also initialize with a full URL
```

src/lib/normalize-url.js
```diff
@@ -0,0 +1,40 @@
...
+// but they usually don't, and we the use the 
+// domain given by the stack through a DOM node
+// in the HTML page (depite it not being a full URL).
```
ptbrowne:
```suggestion
// in the HTML page (despite it not being a full URL).
```

src/lib/normalize-url.js
```diff
@@ -0,0 +1,40 @@
...
+//
+// This function normalize these different cases
```
ptbrowne:
```suggestion
// This function normalizes these different cases
```

src/lib/normalize-url.js
```diff
@@ -0,0 +1,40 @@
...
+// and give a full URL object in response.
```
ptbrowne:
```suggestion
// and returns a URL object.
```

src/lib/normalize-url.js
```diff
@@ -0,0 +1,40 @@
...
+// 
+// The `secure` parameter, if given, always has the
+// priority over the one from the URL.
+// We use a secure protocol by default if we can't decide.
+
```
ptbrowne: JSDoc style comments are picked up by automatic documentation tools are the norm in our code.

Example

```
  /**
   * Creates a {@link DocumentCollection} instance.
   *
   * @param  {String} doctype The collection doctype.
   * @return {DocumentCollection}
   */
```

src/lib/normalize-url.js
```diff
@@ -0,0 +1,40 @@
...
```
ptbrowne:
```suggestion
// Returns a URL object from a partial URL (given by a human) 
```

src/lib/normalize-url.js
```diff
@@ -0,0 +1,40 @@
...
+// in the HTML page (depite it not being a full URL).
+//
+// This function normalize these different cases
+// and give a full URL object in response.
+// 
+// The `secure` parameter, if given, always has the
+// priority over the one from the URL.
+// We use a secure protocol by default if we can't decide.
+
+export default function normalizeURL(cozyURL, ssl) {
```
ptbrowne:
```suggestion
export default function normalizeURL(cozyURL, secure = true) {
```

makes it clearer that we default to true for the secure parameter.

I would rename ssl to secure since https can be over ssl or tls.

src/lib/stack-client.js
```diff
@@ -0,0 +1,234 @@
...
+
+const errorStatuses = {
+  '401': UnauthorizedStackException,
+  '403': ForbiddenException,
+  '404': NotFoundException,
+  '405': MethodNotAllowedException,
+  '500': ServerErrorException
+}
+
+// Utilizy to fetch the resources with cozy-client
```
ptbrowne:
```suggestion
// Utility to fetch resources with cozy-client
```

src/lib/stack-client.js
```diff
@@ -0,0 +1,234 @@
...
+// maintening the compatibility with the legacy
```
ptbrowne:
```suggestion
// maintaining compatibility with the legacy 
```

src/lib/realtime.js
```diff
@@ -0,0 +1,40 @@
...
+  try {
+    realtime
+      .subscribe(realtimeConfig, 'io.cozy.apps')
+      .onCreate(async app => {
+        // Fetch direclty the app to get attributes `related` as well.
+        let fullApp
+        try {
+          fullApp = await getApp(app.slug)
+        } catch (error) {
+          throw new Error(`Cannont fetch app ${app.slug}: ${error.message}`)
```
cedricmessiant: Typo: `Cannont`

src/lib/stack-client.js
```diff
@@ -0,0 +1,364 @@
...
+
+/**
+ * Fetch a resource with cozy-client
+ *
+ * Utility to maintain the compatibility with the legacy 
+ * standalone cozy-bar client
+ *
+ * @function
+ * @private
+ * @returns {Promise} the full raw JSON playload
```
cedricmessiant: Typo: `playload` -> `payload` (multiple times in the file)

src/lib/icon.js
```diff
@@ -9,9 +9,20 @@ const mimeTypes = {
   svg: 'image/svg+xml'
 }
 
+/**
```
ptbrowne: This could be replaced in the future by StackClient.getIconURL with { priority: 'registry' }

src/lib/icon.js
```diff
@@ -9,9 +9,20 @@ const mimeTypes = {
...
+ * Get an icon URL usable in the HTML page from it's stack path
+ * 
+ * @function
+ * @private
+ * @param {function} iconFetcher - takes an icon path on the stack
+ *                                 and returns a fetch object with the icon
+ * @param {object} app - app obbject with a `links.icon` attribute
```
ptbrowne: s/obbject/object

src/lib/icon.js
```diff
@@ -9,9 +9,20 @@ const mimeTypes = {
   svg: 'image/svg+xml'
 }
 
+/**
+ * Get an icon URL usable in the HTML page from it's stack path
+ * 
+ * @function
+ * @private
+ * @param {function} iconFetcher - takes an icon path on the stack
+ *                                 and returns a fetch object with the icon
```
ptbrowne: s/fetch object/Response

src/lib/stack.js
```diff
@@ -1,3 +1,66 @@
...
+ *
+ * @private
+ * @TODO We should set it to `undefined`
+ * and throw an error when it is not initialized
+ * Leaving with that default for today
+ * so I do not need to rewrite deeply the tests
+ * (they test some components without initializing
+ * the stack client first, and rely on non-initialized
+ * legacy client behaviour)
+ */
```
ptbrowne: Nit: Seems that your line wrapping setting is very low

src/lib/stack-client.js
```diff
@@ -1,3 +1,29 @@
...
+ * @function
+ * @param {Object}  arg
+ * @param {Object}  arg.cozyClient - a cozy client instance
+ * @param {Function} arg.onCreateApp 
+ * @param {Function} arg.onDeleteApp
+ * @returns {Promise}
+ */
+const init = function(options) {
+  cozyClient = options.cozyClient
+  const legacyOptions = {
```
ptbrowne: `const stackClient = getStackClient()` so that it is not repeated.

src/lib/stack.js
```diff
@@ -44,7 +48,7 @@ const current = function() {
  * @returns {Promise}
  */
 const init = function(options) {
-  stack = internal
+  stack = (options.cozyClient) ? client : internal
```
ptbrowne: I would find it clearer if it was written

`stack = (options.cozyClient) ? newStack : legacyStack`

test/lib/stack.spec.js
```diff
@@ -0,0 +1,107 @@
...
+    
+    internalInit = internal.init
+    internal.init = jest.fn( () => Promise.resolve() )
+    internalGetApp = internal.get.app
+    internal.get.app = jest.fn( () => Promise.resolve() )
+
+    clientInit = client.init
+    client.init = jest.fn( () => Promise.resolve() )
+    clientGetApp = client.get.app
+    client.get.app = jest.fn( () => Promise.resolve() )
```
ptbrowne: Instead of saving references to the functions you are replacing with mocks, you can use 

```
jest.spyOn(client, 'init').mockImplementation(jest.fn())
````

which does the saving for you and provides you with an easy way to restore things afterwards : in the `afterAll` you can write `jest.restoreAllMocks()`

File: src/lib/stack-client.js [hunk 23]
```diff
@@ -1,11 +1,48 @@
...
+ * @function
+ * @returns {Object} cozy-stack-client instance
+ */
+const getStackClient = function() {
+  return cozyClient.getStackClient()
+}
+
+/**
+ * Logout and disconnect the user
+ * @function
```
ptbrowne: Please add TODO move this to cozy-stack-client.

src/lib/stack-client.js
```diff
@@ -1,11 +1,48 @@
...
+ * @returns {Promise}
+ */
+const logout = function() {
+  return getStackClient().fetch('DELETE', '/auth/login').then(
```
ptbrowne:
```
const logout = async () => {
  const resp = await getStackClient().fetch('DELETE', '/auth/login')
  ....
}
```

lets you get rid of the callback etc...

It will also be more secure in the sense that if the getStackClient() fails synchronously, it will still return a Promise that will reject instead of throwing a synchronous exception.

test/lib/stack-client/stack-client.logout.spec.js
```diff
@@ -0,0 +1,62 @@
...
+let internalLogout
+
+describe("stack client", () => {
+
+  describe("logout", () => {
+
+    const stackClient = {
+      token: { token: "mytoken"},
+      uri: "https://test.mycozy.cloud",
+      fetch: jest.fn(
```
ptbrowne: `jest.fn().mockResolvedValue({ status: 200 })` is a tad bit shorter ;)

test/lib/stack-client/stack-client.logout.spec.js
```diff
@@ -0,0 +1,62 @@
...
+      expect( internal.logout ).not.toHaveBeenCalled()
+    })
+
+    it("should have called cozy-client", () => {
+      expect( cozyClient.getStackClient().fetch ).toHaveBeenCalled()
+    })
+
+    it("should not throw", async () => {
+      expect( () => stack.logout() ).not.toThrow()
+      const thrown = await stack.logout().then(
```
ptbrowne: Maybe you can use the `await expect(stack.logout()).rejects.toBe(undefined)` construct ?

test/lib/stack-client/stack-client.logout.spec.js
```diff
@@ -0,0 +1,62 @@
...
+
+    it("should not forward to the old internal client", () => {
+      expect( internal.logout ).not.toHaveBeenCalled()
+    })
+
+    it("should have called cozy-client", () => {
+      expect( cozyClient.getStackClient().fetch ).toHaveBeenCalled()
+    })
+
+    it("should not throw", async () => {
```
ptbrowne: should not throw when ?

src/lib/stack-client.js
```diff
@@ -65,7 +85,7 @@ const init = function(options) {
   cozyClient = options.cozyClient
   const legacyOptions = {
     ...options,
-    cozyURL: cozyClient.getStackClient().uri,
+    cozyURL: getCozyURLOrigin(),
```
ptbrowne: getCozyURLOrigin could take the `stackClient` so that the call to getStackClient is not duplicated.

test/lib/stack-client/stack-client.cozyurl.spec.js
```diff
@@ -23,10 +23,10 @@ describe("stack client", () => {
       onDeleteApp: function() {},
     }
 
-    beforeAll(() => {
+    beforeAll(async () => {
```
ptbrowne: This commit should be fixed up in the commit that introduces the error.

src/lib/stack-client.js
```diff
@@ -71,6 +83,64 @@ const updateAccessToken = function(token) {
...
+ * Fetch a resource with cozy-client
+ *
+ * Utility to maintain the compatibility with the legacy 
+ * standalone cozy-bar client
+ *
+ * @function
+ * @private
+ * @returns {Promise} the full raw JSON playload
+ */
+const fetchJSON = function(method, path, body, options={}) {
```
ptbrowne: This function should be async so that even if there is a throw in the code building the options, the function still returns a Promise (rejected) and not a synchronous error.

src/lib/stack-client.js
```diff
@@ -141,6 +141,30 @@ const getApps = function () {
...
+ * whole JSON playload from the server
+ *
+ * @function
+ * @param {string} slug 
+ * @returns {Promise} 
+ */
+const getApp = function(slug) {
+  if (!slug) {
+    throw new Error('Missing slug')
+  }
```
ptbrowne: Maybe have this function async ?

IMHO, instead of throwing, a Promise.reject() would be better since the normal return type is a promise and the calling code will expect a promise here. Anyway if the calling code is using await it will work.

src/lib/stack-client.js
```diff
@@ -121,6 +121,51 @@ const fetchJSON = function(method, path, body, options={}) {
...
+      return fn().then(result => 
+        cache = result
+      ).catch( error => 
+        cache = is404(error) ? defaultValue : undefined
+      ) 
+    } else {
+      return Promise.resolve(cache)
+    }
+  }
+}
```
ptbrowne: IMHO the cache should be limited to the cozy you are on. It can be achievied by having the cache as an object and using the stackClient.uri as the key to the cache.

ptbrowne: lodash memoize comes to mind.

```
const key = () => getStackClient().uri
const context = memoize(x => undefinedIf404(getContext()), key)
```

src/lib/stack-client.js
```diff
@@ -210,6 +210,41 @@ const getApp = function(slug) {
...
+ *
+ * When no quota is returned by the server
+ * the quota used is the larger between 
+ * `defaultQuota` and 10 * usage
+ * 
+ * @function
+ * @returns {Object} {usage, quota, isLimited}
+ */
+const getStorageData = function() {
+  return fetchJSON('GET', '/settings/disk-usage').then(
```
ptbrowne: async/await ?

test/lib/stack-client/stack-client.cozyurl.spec.js
```diff
@@ -38,7 +38,7 @@ describe("stack client", () => {
     })
 
     it("should not forward to the old internal client", () => {
-      stack.get.cozyURL
+      stack.get.cozyURL()
```
ptbrowne: Should be fixed up.

test/lib/stack-client/stack-client.cozyfetchjson.spec.js
```diff
@@ -0,0 +1,53 @@
...
+describe("stack client", () => {
+
+  describe("cozyFetchJSON", () => {
+
+    let json = jest.fn(() => {
+      return { 
+        data: { id: "myid" }
+      } 
+    })
+    let fetch = jest.fn((method, path, body, options) => {
```
ptbrowne: mockResolvedValue

src/lib/icon.js
edas: yep. Will see that later. I don't want to mix two different refactoring task

src/lib/stack-client.js
edas: The last stage will only have one call to `getStackClient()` so the repetition is only visible between commits of this PR

src/lib/stack.js
edas: :+1: but it is temporary and disapear at the last stage of this PR anyways

test/lib/stack.spec.js
edas: Oh! 🎉

test/lib/stack-client/stack-client.logout.spec.js
edas: hum… I did not know about `resolves` and `rejects`. `resolves` seems the one that I should use (it will fail if the promise rejects)

src/lib/stack-client.js
edas: it was not true at stage 2 (getCozyURLOrigin did not exists) and won't be true at stage 12 (legacy options do not exist anymore). I'd prefer not to change too many temporary things

src/lib/stack-client.js
edas: Is there a risk to have an execution going through multiple instance ? I don't see how

test/lib/stack-client/stack-client.cozyurl.spec.js
edas: done in a later fix commit

src/lib/stack-client.js
```diff
@@ -0,0 +1,364 @@
...
+ * @param {Function} fn - the function to memoize. It will be
+ *                        called without any parameter
+ * @param {Object} defaultValue - returned in case of 404
+ */                  
+const withCache = function(fn, defaultValue) {
+  let cache = undefined;
+  return function() {
+    if (cache === undefined) {
+      return fn().then(result => 
+        cache = result
```
nono: Isn't a `return` missing here? You save the result in the cache, but the result is no longer available to the called of `withCache`.

src/lib/stack-client.js
```diff
@@ -0,0 +1,364 @@
...
+      ).catch( error => 
+        cache = is404(error) ? defaultValue : undefined
```
nono: Same here

src/lib/stack-client.js
nono: And testing `withCache` would be a good idea. It's a simple function to test, but having the right implementation on the first try is harder.

src/lib/stack-client.js
```diff
@@ -0,0 +1,364 @@
...
+ * cached and reused in the next call to the function.
+ * 
+ *
+ * @function
+ * @param {Function} fn - the function to memoize. It will be
+ *                        called without any parameter
+ * @param {Object} defaultValue - returned in case of 404
+ */                  
+const withCache = function(fn, defaultValue) {
+  let cache = undefined;
```
nono: I'm surprised that no linter told you to remove the final `;`

src/lib/stack-internal.js
```diff
@@ -0,0 +1,201 @@
...
+  }
+  return fetchJSON(`${COZY_URL}${path}`, requestOptions).then(json => {
+    const responseData = Object.assign({}, json.data)
+    if (responseData.id) responseData._id = responseData.id
+    return Promise.resolve(responseData)
+  })
+}
+
+function getStorageData() {
+  return fetchJSON(`${COZY_URL}/settings/disk-usage`, fetchOptions())
```
nono: I'm curious: does it work? It sending `Accept: application/json` instead of the [documented](https://docs.cozy.io/en/cozy-stack/settings/#disk-usage) `Accept: application/vnd.api+json` but the stack often accepts both.

src/lib/stack-internal.js
```diff
@@ -0,0 +1,201 @@
...
+  return () => {
+    return cache['context']
+      ? Promise.resolve(cache['context'])
+      : fetchJSON(`${COZY_URL}/settings/context`, fetchOptions())
+          .then(context => {
+            cache['context'] = context
+            return context
+          })
+          .catch(error => {
+            if (error.status && error.status === 404) cache['context'] = {}
```
nono: We should return something here, no?

src/lib/stack.js
```diff
@@ -1,231 +1,70 @@
...
+
+/**
+ * Get the current stack client (legacy or cozy-client based)
+ * based on which one has been initialized
+ *
+ * @returns {Object} functions to call the stack
+ */
+const current = function() {
+  if (stack === undefined) {
+    throw new Error("client not initialized in cozy-bar")
```
nono: Can we have this case, as `stack` is initialized to `internal`?

src/lib/stack-client.js
```diff
@@ -0,0 +1,364 @@
...
+/**
+ * Logout and disconnect the user
+ * @function
+ * @returns {Promise}
+ */
+const logout = function() {
+  return getStackClient().fetch('DELETE', '/auth/login').then(
+    resp => {
+      if (resp.status === 401) {
+        throw new UnauthorizedStackException()
```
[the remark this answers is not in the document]
edas: It probably is. I tried to refactor without touching the external API

src/lib/stack-client.js
```diff
@@ -0,0 +1,364 @@
...
+ * underlying HTTP request returned a 404
+ * then this function returns a default value
+ * 
+ * In the absence of any other error, the result is
+ * cached and reused in the next call to the function.
+ * 
+ *
+ * @function
+ * @param {Function} fn - the function to memoize. It will be
+ *                        called without any parameter
```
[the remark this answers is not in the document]
edas: I ended using async/await so there is no more any constraint on `fn` to return a Promise

src/lib/stack-client.js
edas: I used to not test internal non exported functions.

As long it's used only one time, let's consider it as an internal mechanism (in the legacy code it was merged in the `getContext` function, I only externalized it to be more readable).

If it ever begin to be used more in the future, then let's put it in its own file, with an `export` and a dedicated test

src/lib/stack-internal.js
edas: It does, and as this is legacy, I wouldn't touch it in this PR (but we could do that in another one if you thinks its usefull)

src/lib/stack-internal.js
edas: Oh ! You probably found a bug in the legacy code.

I am not touching this here as it is a separate matter, but it may be something to lookup

src/lib/stack.js
edas: nope, we don't. 
`stack`should not be set as `internal`by default. I did it to keep compatibility with the legacy tests somewhere else in the code base. I choosed to keep this lines and to put a big TODO warning upper in the file.

src/lib/icon.js
edas: in #469

src/lib/stack-internal.js
edas: #470

src/lib/stack-client.js
ptbrowne: On mobile, you can switch cozies

edas: It does not unload and reload the whole code ? I am reluctant to this because it changes the behaviour of the legacy client

ptbrowne: No in Banks, we do not reload the page after the logout.

edas: ok -> #471

edas:
[x] Code is tested in a sample dev app with a legacy client
[x] Code is tested in a sample dev app with a cozy-client instance

I should have answered all major feedback except: 

* Feedback about lib/stack-internal.js (this is untouched code, not part of the added feature)
* Some style about intermediary commits (on line that disapear anyways at the end of this PR)
* Features & bugs not part of this feature: #469, #470 and #471 
* I choosed to leave synchronous exceptions instead of masking them in a promise rejection. Some were already synchronous in the legacy code anyways. I made this choice to reduce the risk for the exception to be silently ignored my mistake. It should have no impact for someone using async/await anyways.
* The legacy code was mostly a then/catch code and not an async/await one. I mostly kept this style for single promises

Please tell me if one of these choices is a problem for you. Otherwise it's probably ok to merge. I plan a simple merge with no squash (all commits should be functional on their own, it will simplify future debug) but this is open to feedback.

Waiting for formal :+1: after this comment before merging

edas:
linting issues fixed

PR #473, opened 2019-03-01
 feat: Allow custom overrides in the theme 

edas:
Objectif : Permettre la surcharge des couleurs lors d'un changement de thème de la barre.

Pour l'instant ce n'est fait que pour le thème 'primary', et uniquement là où il surcharge la vue mobile.

### DIFF
--- a/src/components/Bar.jsx
+++ b/src/components/Bar.jsx
@@ -180,6 +180,7 @@ class Bar extends Component {
     } = this.state
     const {
       theme,
+      themeOverrides,
       barLeft,
       barRight,
       barCenter,
@@ -189,8 +190,17 @@ class Bar extends Component {
       onLogOut,
       userActionRequired
     } = this.props
+
+    const { 
+      primaryColor: pColor, 
+      primaryContrastTextColor: pctColor 
+    } = themeOverrides
+    const pStyle = pColor ? { '--cozBarThemePrimaryColor': pColor } : { }
+    const pctStyle = pctColor ? { '--cozBarThemePrimaryContrastTextColor': pctColor } : { }
+    const themeStyle = { ...pStyle, ...pctStyle }
+
     return (
-      <div className={`coz-bar-wrapper coz-theme-${theme}`}>
+      <div className={`coz-bar-wrapper coz-theme-${theme}`} style={themeStyle}>
         <div id="cozy-bar-modal-dom-place" />
         <div className="coz-bar-container">
           {barLeft || this.renderLeft()}
@@ -229,7 +239,8 @@ class Bar extends Component {
 }
 
 const mapStateToProps = state => ({
-  theme: getTheme(state),
+  theme: getTheme(state).name,
+  themeOverrides: getTheme(state).overrides,
   barLeft: getContent(state, 'left'),
   barRight: getContent(state, 'right'),
   barCenter: getContent(state, 'center'),
--- a/src/lib/reducers/theme.js
+++ b/src/lib/reducers/theme.js
@@ -2,22 +2,43 @@ const SET_THEME = 'SET_THEME'
 const DEFAULT_THEME = 'default'
 const PRIMARY_THEME = 'primary'
 const THEMES = [DEFAULT_THEME, PRIMARY_THEME]
+const EMPTY_OVERRIDES = {}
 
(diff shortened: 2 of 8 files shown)


src/lib/reducers/theme.js
```diff
@@ -2,22 +2,40 @@ const SET_THEME = 'SET_THEME'
...
+ */
+export const setTheme = (theme=null, overrides=EMPTY_OVERRIDES) => ({
   type: SET_THEME,
-  theme
+  theme: { type: theme, overrides }
 })
 
 // reducer
-export const getDefaultTheme = () => DEFAULT_THEME
+export const getDefaultTheme = () => ({type: DEFAULT_THEME, overrides: EMPTY_OVERRIDES})
```
ptbrowne: I wonder if it would not be simpler to have

```
DEFAULT_THEME = { type: 'default', overrides: {} }
```

not to rebuild the theme object by hand when we need it (also on L38)

ptbrowne: Also `name` could be more appropriate than `type` here since we already have `type` in the action and it can be confusing.

src/lib/reducers/theme.js
```diff
@@ -2,22 +2,40 @@ const SET_THEME = 'SET_THEME'
...
 
 export const reducer = (state = getDefaultTheme(), action) => {
```
ptbrowne: Maybe add a comment on structure of the state. name of the theme and overrides.

src/components/Bar.jsx
```diff
@@ -189,8 +190,17 @@ class Bar extends Component {
...
       userActionRequired
     } = this.props
+
+    const { 
+      primaryColor: pColor, 
+      primaryContrastTextColor: pctColor 
+    } = themeOverrides
+    const pStyle = pColor ? { '--cozBarThemePrimaryColor': pColor } : { }
+    const pctStyle = pctColor ? { '--cozBarThemePrimaryContrastTextColor': pctColor } : { }
+    const themeStyle = { ...pStyle, ...pctStyle }
```
ptbrowne:
```
    const pStyle = pColor ? { '--cozBarThemePrimaryColor': pColor } : null
    const pctStyle = pctColor ? { '--cozBarThemePrimaryContrastTextColor': pctColor } : null
    const themeStyle = pStyle || pctStyle ? { ...pStyle, ...pctStyle } : null
```

This way, when pColor and and pctStyle is empty, we do not put nothing in style. See the snaps below with lots of empty `style`.

src/lib/reducers/theme.js
edas: :+1: added upper

drazik:
With this implementation (if I understand well), we need to override JS files that call `cozy.bar.setTheme('primary')` to replace it with `cozy.bar.setTheme('primary', overrides)`. I think it's too many efforts to achieve what we want.

I think we can to it with only CSS variables and nothing on the JS side. What about something like this:

```css
[role=banner] .coz-bar-wrapper
    box-shadow inset 0 -1px 0 0 var(--silver)

    .coz-nav-apps-btns
        color var(--slateGrey)

    .coz-bar-burger
        color var(--coolGrey)

@media (max-width: 64em)
    [role=banner] .coz-bar-wrapper
        background-color var(--white)

@media (max-width: 48em)
    [role=banner] .coz-bar-wrapper
        &.coz-theme-primary
            box-shadow inherit
            background-color var(--cozBarThemePrimaryColor, var(--primaryColor), <some last fallback color>)

            .coz-nav-apps-btns,
            .coz-bar-burger
                color var(--cozBarThemePrimaryContrastTextColor, var(--primaryContrastTextColor), <some last fallback color>)
```

The on the app side, we would just have to define `--cozBarThemePrimaryColor` and `--cozBarThemePrimaryContrastTextColor` if we want a special primary color for the cozy-bar.

edas:
vu ensemble sur zoom. L'objectif de @drazik était d'éviter une duplication de fichier dans l'app appelante, et on a trouvé une solution

drazik:
Yes. Voici la PR sur Banks : https://github.com/cozy/cozy-banks/pull/974

Si c'est OK pour vous côté Banks, alors on peut merger cette PR.

Issue #172, opened 2018-02-21
babel-polyfill ?


y-lohse:
Hi, I found a problem and I think the fix should be in cozy-bar, but let's discuss this.

### The problem

On browsers that don't support `Object.assign` (which includes IE11, which we support), some apps crash with the error `Object doesn't support property or method 'assign'`.
More specifically, this error is emitted by the cozy-bar, when it imports the `Modal` component from cozy-ui which uses [`Object.assign`](https://github.com/cozy/cozy-ui/blob/master/react/Modal/index.jsx#L167) right away.

### the solution

We really want to polyfill `Object.assign`, and in fact, we mostly do. Our apps don't have this problem. But cozy-bar doesn't polyfill it, and I really think it should.
To be clearer, cozy-bar uses the `cozy-app` babel preset, which *replaces* it's own `Object.assign` calls, but does *not* polyfill `Object.assign`. So when cozy-ui tries to use it it crashes. We're supposed to `import 'babel-polyfill'` at the very begining of the app for that.

### The alternative

Noticed that I said only *some* apps crash? A couple aren't affected, because they load the app's code (including the polyfill)  before they load `cozy-bar.js`. They tend to use [this form](https://github.com/cozy/cozy-settings/blob/master/src/index.ejs#L11) of templating.
So we could also switch all apps to this format.

___________

If we import `babel-polyfill` in `cozy-bar`, we'll essentially load it twice on each page — once in the app and once in the bar. Which makes sense because both are standalone files, but it's also annoying.

Personally I think it's better to take the weight hit, and be less fragile. Third party devs won't know they have to provide their own polyfills and might run into the same problem.

goldoraf:
That's yet another reason to stop serving the bar by the stack, we should include it in our apps like a component.

y-lohse:
Just a head's up, this is breaking apps in production. i'll add the polyfills to cozy-bar on monday, unless someone has a strong objection (and a better plan).

CPatchane:
Hi @y-lohse, thanks for this issue. The problem should already be fixed in the [last version of `babel-preset-cozy-app`](https://github.com/CPatchane/create-cozy-app/blob/master/CHANGELOG.md#create-cozy-app055-cozy-scripts051-cozy-scripts-vanilla015-babel-preset-cozy-app032-eslint-config-cozy-app032-2018-02-16). I think we just need to update the dependancy in the `cozy-bar`, don't we?

ptbrowne:
[...]

PR #476, opened 2019-03-05
fix: Stop polyfiling Object.assign

y-lohse:
We import a polyfill for `Object.assign` from `core-js`. Since version 2.6.4, this somehow also polyfills `fetch`, which in turn causes other problems.

The initial reason for using the polyfill is discussed [here](https://github.com/cozy/cozy-bar/issues/172) but it boils down to supporting `Object.assign` calls in cozy-ui on IE11. In the meantime, we transpile cozy-ui so `Object.assign` calls are transformed when needed, but also we dropped IE11 support. So I suggest we remove this dependency altogether from the bar.

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,6 @@
     "@babel/core": "7.2.2",
     "babel-core": "7.0.0-bridge.0",
     "babel-preset-cozy-app": "1.3.2",
-    "core-js": "2.6.4",
     "cozy-client-js": "0.14.2",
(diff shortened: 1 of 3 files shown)


y-lohse:
[upgrade Cozy-bar & problème IDB](https://trello.com/c/n88JvtDE/1698-upgrade-cozy-bar-probl%C3%A8me-idb)

PR #484, opened 2019-03-15
Minor revamp

CPatchane:
Minor stack refactoring for better testing and code understanding

### DIFF
--- a/src/lib/stack.js
+++ b/src/lib/stack.js
@@ -65,5 +65,8 @@ export default {
   get,
   updateAccessToken: (...args) => current().updateAccessToken(...args),
   logout: (...args) => current().logout(...args),
-  cozyFetchJSON: (...args) => current().cozyFetchJSON(...args)
+  cozyFetchJSON: (...args) => current().cozyFetchJSON(...args),
+  // useful to connect some getters outside of this file without exposing
+  // directly the private stack variable
+  getStack: current
 }
--- a/src/lib/stack.js
+++ b/src/lib/stack.js
@@ -6,15 +6,8 @@ import client from 'lib/stack-client'
  * on which one has been initialized
  *
  * @private
- * @TODO We should set it to `undefined`
- * and throw an error when it is not initialized
- * Leaving with that default for today
- * so I do not need to rewrite deeply the tests
- * (they test some components without initializing
- * the stack client first, and rely on non-initialized
- * legacy client behaviour)
  */
-let stack = internal
+let stack
 
 /**
  * Get the current stack client (legacy or cozy-client based)
@@ -23,7 +16,7 @@ let stack = internal
  * @returns {Object} functions to call the stack
  */
 const current = function() {
-  if (stack === undefined) {
+  if (!stack) {
     throw new Error('client not initialized in cozy-bar')
   }
   return stack
--- a/test/lib/stack.spec.js
+++ b/test/lib/stack.spec.js
@@ -2,18 +2,16 @@ import internal from 'lib/stack-internal.js'
 import client from 'lib/stack-client.js'
 import stack from 'lib/stack.js'
 
-
-const cozyURL = "https://test.mycozy.cloud"
-const token = "mytoken"
+const cozyURL = 'https://test.mycozy.cloud'
+const token = 'mytoken'
 const onCreateApp = function() {}
 const onDeleteApp = function() {}
 
-describe("stack proxy", () => {
-
+describe('stack proxy', () => {
   beforeAll(() => {
     jest.spyOn(internal, 'init').mockResolvedValue(undefined)
     jest.spyOn(internal.get, 'app').mockResolvedValue(undefined)
-    
+
     jest.spyOn(client, 'init').mockResolvedValue(undefined)
     jest.spyOn(client.get, 'app').mockResolvedValue(undefined)
   })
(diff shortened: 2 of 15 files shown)


test/lib/stack.spec.js
```diff
@@ -22,66 +20,76 @@ describe("stack proxy", () => {
...
+    it('should call the client stack init', () => {
+      expect(client.init).toHaveBeenCalled()
     })
 
-    it("should forward requests to the cozy-client stack client", () => {
-      expect( client.get.app ).toHaveBeenCalled()
+    it('should forward requests to the cozy-client stack client', () => {
+      expect(client.get.app).toHaveBeenCalled()
     })
-
```
ptbrowne: It seems this commit also contains linting, it would be nice to separate the two commits.

src/lib/stack.js
```diff
@@ -65,5 +65,7 @@ export default {
   get,
   updateAccessToken: (...args) => current().updateAccessToken(...args),
   logout: (...args) => current().logout(...args),
-  cozyFetchJSON: (...args) => current().cozyFetchJSON(...args)
+  cozyFetchJSON: (...args) => current().cozyFetchJSON(...args),
+  // allow to connect other getters
```
ptbrowne: Allows

I have trouble understanding the comment, it could be more detailed IMHO.

ptbrowne:
I don't understand well what is going on, it does not seem to me like a minor revamp (+248-536), could you explain in more details what is the goal, the initial problem etc..

CPatchane:
It's minor since it's just about cleaning and simplifying some part of the code for better testing, understanding and to make customisation easier.

test/lib/stack.spec.js
edas: 👍 on this for the next time.

test/lib/stack.spec.js
```diff
@@ -22,66 +20,76 @@ describe("stack proxy", () => {
...
+  it('should throw error if trying to use the stack before init', () => {
+    jest.clearAllMocks()
+    const params = { cozyURL, token, onCreateApp, onDeleteApp }
+
+    // TODO for now the stack is default to internal even before init
+    // expect(() => {
+    //   stack.getStack()
+    // }).toThrowErrorMatchingSnapshot()
+    stack.init(params)
+    expect(() => {
```
edas: You probably don't need the check for exceptions as long as you check for marching Snapshots (if there is an exception, your snapshot won't match anyways)

test/lib/stack.spec.js
```diff
@@ -22,66 +20,76 @@ describe("stack proxy", () => {
...
+      expect(client.get.app).toHaveBeenCalled()
     })
-    
   })
 
+  it('should throw error if trying to use the stack before init', () => {
+    jest.clearAllMocks()
+    const params = { cozyURL, token, onCreateApp, onDeleteApp }
+
+    // TODO for now the stack is default to internal even before init
```
edas: Why are you leaving these commented lines ?

test/lib/stack.spec.js
```diff
@@ -22,66 +20,76 @@ describe("stack proxy", () => {
...
+    // expect(() => {
+    //   stack.getStack()
+    // }).toThrowErrorMatchingSnapshot()
+    stack.init(params)
+    expect(() => {
+      stack.getStack()
+    }).not.toThrowError()
+    expect(stack.getStack()).toMatchSnapshot()
```
edas: Pourquoi compares-tu un snapshot ? Ca ne correspond pas au test décrit. Que cherches-tu à vérifier / confirmer ?

CPatchane: Je voulais vérifier le fait que la stack est correctement défini et j'ai jugé plus malin de placé un snapshot qui permet aussi de vérifier qu'il a les bonnes propriétés au passage, mais je me trompe peut être un est `.toBeDefined()` pourrait suffire.

test/lib/stack.spec.js
CPatchane: It could still throw an exception after the init for some reasons, so I check that after the init, using the stack won't throw any exception
Or maybe I didn't understand your point here

CPatchane:
This PR just cleans and revamps some little parts of the code without changing the way the Cozy-bar works to make customisation easier and for better testing by the way. And there won't be any future PR after this one (at least about this part of the bar) @edas 
Even for the fixed todo, we're all initialising the cozy-bar correctly before using it. Only some Bar tests were using the Bar components without stack initialisation and this is fixed here.

test/lib/stack.spec.js
edas: If `stack.getStack()` is throwing an exception in L91, it will also do it in  any code is throwing an exception, it will also do it in L93 so the test will fail anyways in L93.

I don't see any case where the L90-92 is adding any protection not found elsewhere. Am I wrong ?

test/lib/stack.spec.js
edas: Le snapshot c'est un peu l'arme atomique. Ca fait des tests qui vont casser à la moindre amélioration. Si ton test c'est "vérifier que le prototype n'a pas changé" alors vas-y pour un snapshot.

Si le sujet c'est 'should throw error if trying to use the stack before init' alors tu vérifies que ça ne jette pas d'erreur et c'est tout.

Là tu as un test qui dit "je vérifie que ça ne jette pas d'erreur si j'utilise la stack avant l'init" mais : 
- Tu vérifies l'inexistance d'erreur *après* l'init
- Tu vérifies que le résultat de l'init reste constant

Bref, ça ne correspond pas.

CPatchane: D'accord, je supprime cette partie pour que ça corresponde au test du coup

test/lib/stack.spec.js
gregorylegarec: The difference is that expecting the call to not throw an execption will make the tests reject, but not fail.

The result is almost the same, without `expect().not.toThrow()` test will "explode" with a dirty Error.
With `expect().not.toThrow()`, the Error will be cleanly caught and the test will reject but with a pretty and nice message.

edas: 👌

CPatchane:
[(𝐌) [BAR] Afficher la liste des connecteurs installés dans la cozy-barre Orange](https://trello.com/c/7Q7WAbUP/1055-%F0%9D%90%8C-bar-afficher-la-liste-des-connecteurs-install%C3%A9s-dans-la-cozy-barre-orange)

PR #486, opened 2019-03-18
Correctly use sprite loader

CPatchane:
Avoid using webpack loader with the inline way
It's also not used in css anymore since a while, so we can remove it.


CPatchane:
[(𝐌) [BAR] Afficher la liste des connecteurs installés dans la cozy-barre Orange](https://trello.com/c/7Q7WAbUP/1055-%F0%9D%90%8C-bar-afficher-la-liste-des-connecteurs-install%C3%A9s-dans-la-cozy-barre-orange)

PR #490, opened 2019-03-19
Make excludes apps list setting up by config file

CPatchane:
Make excludes apps list setting up by a config file, will allow to customise this list

### DIFF
--- a/src/config/excludedApps.yaml
+++ b/src/config/excludedApps.yaml
@@ -1,0 +1,1 @@
+- settings
--- a/src/lib/reducers/apps.js
+++ b/src/lib/reducers/apps.js
@@ -1,5 +1,6 @@
 import stack from 'lib/stack'
 import unionWith from 'lodash.unionwith'
+import EXCLUDES from 'config/excludedApps'
 
 // constants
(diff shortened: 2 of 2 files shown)


CPatchane:
[(𝐒) [BAR] ETUO j'ai accès à Settings via la liste des applications](https://trello.com/c/gmAAk8wN/1060-%F0%9D%90%92-bar-eto-jai-acc%C3%A8s-%C3%A0-settings-via-la-liste-des-applications)

src/config/excludedApps.yaml
```diff
@@ -0,0 +1 @@
+- settings
```
nono: Why a YAML and not a JSON file?

gregorylegarec:
I guess we can just override this configuration file to customize or Cozy-Bar ? If so, LGTM.

Same question as @nono by the way. Why YAML ?

src/config/excludedApps.yaml
CPatchane: We were already using those files for configs in cozy-bar so, it allows to add comments in configuration even if for here YAML or JSON doesn't matter

PR #494, opened 2019-03-20
Inject cozyClient in React context

gregorylegarec:
Inject cozyClient into children components like in any other app.

CozyClient is now accessible in any child via

```
const cozyClient = this.context.client
```

Query component can be used, as well as HOC withQuery or withMutations.


ptbrowne:
It looks like the Bar would be more easily used as React component.

```
const App = () => {
  return <Layout>
    <Bar />
   </Layout>
}
```

It would then "inherit" from the cozyClient without this.

I think if you agree that we should keep this objective in mind while we refactor.

edas:
> It looks like the Bar would be more easily used as React component.

:+1: for this in the future

package.json
```diff
@@ -87,10 +87,11 @@
     "@babel/core": "7.3.4",
     "babel-core": "7.0.0-bridge.0",
     "babel-preset-cozy-app": "1.3.2",
+    "cozy-client": "^6.15.0",
     "cozy-client-js": "0.14.2",
     "cozy-device-helper": "1.6.10",
     "cozy-interapp": "0.4.4",
-    "cozy-realtime": "2.0.6",
+    "cozy-realtime": "2.0.7",
```
Crash--: not needed is this pr ?

gregorylegarec:
@CPatchane What do we do about this ?

```
 FAIL  ./dist/cozy-bar.min.js: 230.93KB > maxSize 174KB (gzip) 
```

gregorylegarec:
@edas The only test available for the index file is failing and is disabled. It looks pretty time consuming to fix this and setup a new test for this condition right now.

As it is "just" a condition and it's already working on my dev environment, I would suggest to delay the implementation of this test, especially if we move forward quickly to provide a Bar as a component, as @ptbrowne is suggesting.

edas:
> @edas The only test available for the index file is failing and is disabled. 

I have mixed feelings about this. Let's go today as is but please, consider that tests *should* be done each time you add or fix a feature

In all case, having a failed test disabled bugs me. What's your plan (outside of this PR) to fix the test? What is the problem?

gregorylegarec:
@edas I've just discoverd this disabled test as I was aiming to set up a test. So it bugs me too and I will ask @CPatchane why we are in this situation.

Btw, I agree, we must always provide tests.

CPatchane:
@gregorylegarec I can't say anything else for now except increase the max size of bundlesize to fix the build here :/
@gregorylegarec @edas A previous PR disabled this test, also to be fixed later :/ https://github.com/cozy/cozy-bar/pull/254#discussion_r220458418

gregorylegarec:
@edas @CPatchane I plead guilty on this one. https://github.com/cozy/cozy-bar/commit/6c2f634670e5cb3d20c3340b78823b31dd91317b#diff-faa503f61f8addeb603095cc2f40c82eL8

gregorylegarec:
Importing `cozy-client` bring us 57 more KB in the bar.

I think we should be able to import this component separately.

ptbrowne:
What do you mean by separately ?

CozyClient is transpiled without `presetEnv:{modules: false}`, thus `imports` are transformed into `require`, which means webpack cannot tree shake.

Can you try adding `presetEnv:{modules: false}` to cozy-client ?

nono:
I think we also need to use `import` and not `require` to import the client in the code of cozy-bar to not block the tree shaking.

gregorylegarec:
@CPatchane Do we agree that we are stuck using `require` until all our apps have migrated to `React` ?

CPatchane:
Yep, we need to use `require` here for all libraries which are using a React library internally :/

gregorylegarec:
Settings `presetEnv:{modules: false}` in cozy-client did not change the final size of the bar.

PR #496, opened 2019-03-25
feat: Use deeplink for the web version of the bar

Crash--:
The idea is to open from the "mobile web app" the native app if installed. If not, we redirect to the web version.

### DIFF
--- a/src/components/Apps/AppItem.jsx
+++ b/src/components/Apps/AppItem.jsx
@@ -6,7 +6,9 @@ import {
   checkApp,
   startApp,
   isAndroidApp,
-  isMobileApp
+  isMobileApp,
+  isMobile,
+  openDeeplinkOrRedirect
 } from 'cozy-device-helper'
 import expiringMemoize from 'lib/expiringMemoize'
 import AppIcon from 'cozy-ui/react/AppIcon'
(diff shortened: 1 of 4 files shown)


Crash--:
[(𝐒) Cozy-bar & deeplink](https://trello.com/c/zQIRE3tI/1686-%F0%9D%90%92-cozy-bar-deeplink)

src/components/Apps/AppItem.jsx
```diff
@@ -99,6 +101,17 @@ export class AppItem extends React.Component {
       // target app is web
       // run switch listener only if the current app is native mobile
       onClick = this.onAppSwitch
+    } else if (isMobile()) {
```
CPatchane: What is the difference with the `isMobileApp()` call?

Crash--: `isMobileApp` checks the cordova context aka the native one. 

Here, we just test if it's the mobile web version... Thanks cozy-device-helper and its name :/

CPatchane: It should be more explicit on cozy-device-helper, but I won't block this PR for that

y-lohse: But hang on, `isMobileAppAvailable ` can only be checked when `isMobile()` is `true`, right? So maybe we could make this triple condition a bit clearer?

Crash--: In fact the three tests are: 
- if we are on native app and other native app is installed
- else if we are on native app 
- else if we are on mobile 

How can I make them clearer? By renaming the variables?

y-lohse: Yeah I think renaming the variables or being more explicit in the commend would help. I know it's not part of this PR but `run switch listener` for example is not very clear.

Something like this I think would help:

```
const isNativeToNative = isMobileAppAvailable
const isNativeToWeb = isMobileApp()
const isWebToNative = isMobile()
```

src/components/Apps/AppItem.jsx
```diff
@@ -99,6 +101,17 @@ export class AppItem extends React.Component {
...
+      //If we are on the "mobile web version", we try to open the native app
+      //if it fails, we redirect to the web version of the requested app
+      onClick = e => {
+        if (NATIVE_APP_INFOS[app.slug]) {
```
y-lohse: Nit: this condition might be better with the `if (isMobile())`.

PR #499, opened 2019-03-26
Add parameter to include or not home app in apps list

CPatchane:
Allow to handle home application differently or not


gregorylegarec:
Why can't we use https://github.com/cozy/cozy-bar/blob/master/src/config/excludedApps.yaml to add home ?

If it is a matter of displaying or not the Home application, I think we should use the same logic, and pass a list of displayed apps instead of a flag for a given app.

CPatchane:
It's not about cozy-home here but the application considered as Home in the Cozy, it's why this is handled differently

PR #502, opened 2019-03-28
fix: Check correctly the slug

Crash--:
I had "can't read slug of undefined" because there is no parenthesis here: 
`app.slug !== homeApp && homeApp.slug` 

and if we add them, prettier remove them, so I just declared a const before ^^"


Crash--:
@ptbrowne I don't know why you're saying that... ;) 

I have added : 
- Proptypes 
- Snapshot for all the cases i think

PR #503, opened 2019-03-28
feat: No need to override anymore if it's the same app

Crash--:
We don't need to override the current application since it breaks some use case (coming from deeplink for instance)

### DIFF
--- a/src/components/Apps/AppItem.jsx
+++ b/src/components/Apps/AppItem.jsx
@@ -117,12 +117,6 @@ export class AppItem extends React.Component {
       }
     }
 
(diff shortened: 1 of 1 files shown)


Crash--:
[cozy-bar - app consultée cliquable](https://trello.com/c/JudDXwcl/1815-cozy-bar-app-consult%C3%A9e-cliquable)

CPatchane:
What's happen if we click on the current app now?

Crash--:
We redirect to the « home » of the app (native or web). See the linked trello card the a more precise description ! ;)

PR #507, opened 2019-03-29
 fix: Don't subscribe to realtime if is public

### LINKED ISSUE #506: Cozybar on public page
Since we inject a CozyClient from the app to the bar, if we inject a non identified client (from a public view for instance) , we got an uncaught error : 
```
Uncaught Error: The application can't subscribe to io.cozy.apps
    at WebSocket.onSocketMessage (cozy-bar.js:17464)
```

This error is only blocking in our end to end test since every non catched error generate a real error. 

We currently fixed the error by removing the injection of cozyclient (tmp fix https://github.com/cozy/cozy-drive/pull/1530)

Crash--:
It's a try to fix #506 . I don't know if this is the way to go? Any ideas ?

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -204,7 +204,8 @@ const init = async ({
     token,
     onCreate: data => reduxStore.dispatch(onRealtimeCreate(data)),
     onDelete: data => reduxStore.dispatch(onRealtimeDelete(data)),
-    ssl
+    ssl,
+    isPublic
   })
   if (lang) {
     reduxStore.dispatch(setLocale(lang))
--- a/src/lib/stack-client.js
+++ b/src/lib/stack-client.js
@@ -350,12 +350,11 @@ const getSettingsAppURL = function() {
  * @param {Object}  arg.cozyClient - a cozy client instance
  * @param {Function} arg.onCreate
  * @param {Function} arg.onDelete
- * @param {Boolean} arg.isPublic - if the bar is called from a public context
  * @returns {Promise}
  */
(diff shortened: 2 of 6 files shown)


Crash--:
> Pas forcément fana du nom isPublic mais je n'ai rien de plus clair à proposer à chaud. Peut-être inPublicContext ou notAuthenticated ?

 
Oui je suis d'accord. Si on part sur ce genre de fix, je renommerai la variable un peu partout car elle fait partie de l'API existante déjà

CPatchane:
D'accord aussi pour la renommer en `notAuthenticated` 👍

ptbrowne:
In a public page, does not the cozyClient have a token ? I am trying to know if it would be possible to know if we are in a public page without explicitly telling so. I think cozyClient should be the one knowing this.

Crash--:
CozyClient has a token and I just found out that it also has a `isLogged` attribute (https://github.com/cozy/cozy-client/blob/bbaa3db6f25d7f78aca7a1f51ca1b3d528217c59/packages/cozy-client/src/CozyClient.js#L74) 

But I don't know if I should change all the bar `isPublic` instances to `isLogged` since I don't know if this is the same behavior everywhere. Can we be logged on a public page?

Crash--:
Quelques news de cette PR : 

- Quand on est en "stack-client", je me base sur "isLogged" du cozyclient, afin de ne pas souscrire au real time 
- Quand on est en "stack-internal", je continue de me baser sur le isPublic pour y souscrire

nono:
Ping, can we merge this PR?

CPatchane:
Thanks @Crash-- 👍

PR #517, opened 2019-04-05
fix: AppItem cursor

Crash--:
Since the active AppItem is now clickable, we should not override the cursor type


Crash--:
[cozy-bar - app consultée cliquable](https://trello.com/c/JudDXwcl/1815-cozy-bar-app-consult%C3%A9e-cliquable)

PR #526, opened 2019-04-12
Don't check drive app on app name

CPatchane:
If the Drive app don't have 'Cozy Drive' name, the bar will be hided, the slug is more reliable for this case.


CPatchane:
[🔶 (𝐒) ETUO ma modale de succès me ramène vers mon assistant financier](https://trello.com/c/tt5EvQgw/1238-%F0%9F%94%B6-%F0%9D%90%92-etuo-ma-modale-de-succ%C3%A8s-me-ram%C3%A8ne-vers-mon-assistant-financier)

PR #544, opened 2019-05-07
Allow overriding AppItem href

gregorylegarec:
This PR is a quick try to answer to our need of passing argument in an AppLinker href in Cozy-Bar.

~~The idea is to be able to override current href using Component inheritance:~~

~~Another alternative should be to implement directly query parameter as an `AppLinker` prop and let it generate the href. What do you think ?~~

How this should work:

```jsx
import 'AppItem' from 'original-bar/src/components/AppItem'

AppItem.buildQueryParams = (props, context) => ({
  foo: 'bar'
})

export AppItem
```

### DIFF
--- a/src/components/Apps/AppItem.jsx
+++ b/src/components/Apps/AppItem.jsx
@@ -9,6 +9,17 @@ import stack from 'lib/stack'
 import PropTypes from 'prop-types'
 
 export class AppItem extends React.Component {
+  /**
+   * Used to add query params to AppLinker links, useful in overrides
+   * @param  {Object} props   AppItem props
+   * @param  {Object} context AppItem context
+   * @return {Object}         Query string parameters as object
+   */
+  static buildQueryParams = () => {
+    // default behaviour
+    return null
+  }
+
   constructor(props) {
     super(props)
     this.onAppSwitch = this.onAppSwitch.bind(this)
@@ -16,6 +27,17 @@ export class AppItem extends React.Component {
 
   componentWillUnmount() {
     if (this.switchTimeout) clearTimeout(this.switchTimeout)
+  }
+
+  buildAppUrl(href) {
+    const url = new URL(href)
+    const queryParams = AppItem.buildQueryParams(this.props, this.context)
+    if (queryParams) {
+      for (const name in queryParams) {
+        url.searchParams.append(name, queryParams[name])
+      }
+    }
+    return url.toString()
   }
 
   onAppSwitch() {
@@ -46,7 +68,7 @@ export class AppItem extends React.Component {
             >
               <a
                 role="menuitem"
-                href={href}
+                href={this.buildAppUrl(href)}
                 data-icon={dataIcon}
                 title={label}
                 onClick={onClick}
--- a/test/components/AppItem.spec.jsx
+++ b/test/components/AppItem.spec.jsx
@@ -1,6 +1,6 @@
 import React from 'react'
 import { AppItem } from 'components/Apps/AppItem'
-import { mount } from 'enzyme'
+import { mount, shallow } from 'enzyme'
 import toJson from 'enzyme-to-json'
 import { tMock } from '../jestLib/I18n'
 import { isMobileApp, isMobile } from 'cozy-device-helper'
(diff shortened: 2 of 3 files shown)


Crash--:
Why not just changing the call to <AppItem> ?  
<AppItem app={{href:${app.href}?foo=bar}} ?

CPatchane:
@Crash-- Using a method here avoid to rewrite all the render method

ptbrowne:
- I try to avoid inheritance when possible, and it's clearer to me if it's the component itself that is overrided and not the caller, as we might forget at some point one calling point.
- I like to use static methods that are overrided

```jsx
class AppItem extends React.Component {
  static methodBeggingToBeOverrided(props) {
    return 'hello'
  }

  static methodUsingOverridedAndContainingTheComplicatedLogic() {
    const res = AppItem.methodBeggingToBeOverrided(this.props)
    return doComplexLogic(res)
  }
}
```

This way when overriding, it is easy to do: 

```jsx
import AppItem from 'original-place-where-is-app-item'

AppItem.methodBeggingToBeOverrided = props => { 
  return props
}

export default AppItem
```

Passing props helps in case the overriden logic needs to use something from the context.

```diff
 import AppItem from 'original-place-where-is-app-item'

 AppItem.methodBeggingToBeOverrided = props => {
+   const client = this.props.client
+   // do something with client
   return props
 }

- export default AppItem
+ export default withClient(AppItem)
```

gregorylegarec:
Little update following @ptbrowne suggestions (ping @Crash--)

test/components/AppItem.spec.jsx
```diff
@@ -26,6 +26,38 @@ jest.mock('cozy-device-helper', () => ({
...
+
+    it('should update query string', () => {
+      AppItem.buildQueryParams = () => ({
+        foo: 'bar',
+        bar: 'buz'
+      })
+
+      const wrapper = shallow(<AppItem app={app} t={tMock} />)
+      const url = wrapper.instance().buildAppUrl('http://fake.fr')
+      expect(url.toString()).toBe('http://fake.fr/?foo=bar&bar=buz')
```
ptbrowne: If this fails, buildQueryParams is not put back to the original one. This can lead to false positives for other tests.

One way to mitigate this:

```js
afterEach(() => {
  jest.restoreAllMocks()
})

it('should update query string', () => {
      jest.spyOn(AppItem, 'buildQueryParams').mockImplementation(() => ({
        foo: 'bar',
        bar: 'buz'
      })
     ...rest of the test
})
```

src/components/Apps/AppItem.jsx
```diff
@@ -9,6 +9,17 @@ import stack from 'lib/stack'
 import PropTypes from 'prop-types'
 
 export class AppItem extends React.Component {
+  /**
+   * Returns query string parameters as object
```
ptbrowne:
```suggestion
   * Used to add query params to AppLinker links, useful in overrides
```

src/components/Apps/AppItem.jsx
```diff
@@ -18,6 +29,17 @@ export class AppItem extends React.Component {
...
 
+  buildAppUrl(href) {
+    const url = new URL(href)
+    const queryParams = AppItem.buildQueryParams(this.props, this.context)
+    if (queryParams) {
+      for (const name in queryParams) {
+        url.searchParams.append(name, queryParams[name])
+      }
+    }
+    return url
```
Crash--: It should be url.toString(), no ? Otherwise we will have an object

test/components/AppItem.spec.jsx
gregorylegarec: Thanks, my mistake, it is how I use to proceed too.

test/components/__snapshots__/AppItem.spec.jsx.snap
```diff
@@ -22,7 +22,7 @@ exports[`app icon should render correctly 1`] = `
     >
       <a
         data-icon="icon-cozy-drive"
-        href="http://fake.fr"
+        href={"http://fake.fr/"}
```
Crash--: I knew it!

test/components/AppItem.spec.jsx
```diff
@@ -26,6 +26,38 @@ jest.mock('cozy-device-helper', () => ({
...
+}
+
+describe('AppItem', () => {
+  describe('buildAppUrl', () => {
+    const originalBuildQueryParams = AppItem.buildQueryParams
+
+    it('should return untouched href', () => {
+      const wrapper = shallow(<AppItem app={app} t={tMock} />)
+      const url = wrapper.instance().buildAppUrl('http://fake.fr')
+      expect(url.toString()).toBe('http://fake.fr/')
```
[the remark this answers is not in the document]
gregorylegarec: Yes, tests should have indicated me that something was wrong when I called `toString()` method. 👍

test/components/__snapshots__/AppItem.spec.jsx.snap
gregorylegarec: ;)

ptbrowne: I did not really understand that, why is there the change ?

ptbrowne:
:shipit:

PR #548, opened 2019-05-08
feat: Terminates realtime on client logout

ptbrowne:
The bar inits realtime and attaches handlers on initialization.
It would not unsubscribe handlers on client logout, which means
that a realtime socket would continue to exist after logout.

To fix this problem, initializeRealtime now returns a function
that serves to unsubscribe handlers. When the stack-client calls
initializeRealtime, it also attaches a one-time handler to
terminate the realtime on logout.

Another potential problem was that the bar, if initialized with
a logged out client would never initialize the realtime. To
mitigate this problem: if the bar is initialized with a logged out
client, it attaches a one-time handler on the "login" event to
initialize the realtime.


src/lib/stack-client.js
```diff
@@ -353,13 +353,25 @@ const getSettingsAppURL = function() {
...
+      getApp,
+      onCreate,
+      onDelete,
+      cozyClient
+    })
+  }
+  if (cozyClient.isLogged) {
+    onLogin()
+  } else {
+    cozyClient.once('login', onLogin)
```
Crash--: What is the main reason to have only a once handler? Why not just a regular on('login') ? What happens if this client is not logged then login then logout then login back?

ptbrowne: Fair point. In Banks, we init the bar on `login` from the client. Therefore, the `init` function above is called at each `login`. This means that in banks, the `else` codepath would not be used. I think I am going to remove the commits introducing the once "login". Let's settle on the fact that the bar should be initialized when the client is logged in. WDYT ?

Crash--: In Drive we manually call Bar.init() when it's needed. We display a few onboarding screens after the login, but we don't want to init the bar yet. 

I think this is the responsibility of the App to init the bar when it needs. 

> Let's settle on the fact that the bar should be initialized when the client is logged in.  

Yep.

ptbrowne: FYI, I removed the `once('login', initRealtime)` commit.

src/lib/stack-client.js
```diff
@@ -353,13 +353,23 @@ const getSettingsAppURL = function() {
...
+    terminateRealtime = initializeRealtime({
+      getApp,
+      onCreate,
+      onDelete,
+      cozyClient
+    })
+  }
+  if (cozyClient.isLogged) {
+    onLogin()
+  }
```
edas: Tu ne tentes pas de réagir à l'événement de login s'il arrive plus tard ?

Crash--:
@ptbrowne Les issues dont du parles on clairement besoin d'être fixée. Mais vu qu'un gros boulot a été fait en parallèle sur le realtime et l'intégration dans la bar, est-ce qu'on a encore besoin de cette PR ?

ptbrowne:
> Mais vu qu'un gros boulot a été fait en parallèle sur le realtime et l'intégration dans la bar

Tu parles de quoi en particulier ?

ptbrowne:
Ca me dérange pas de rouvrir une issue "Faire de la bar un plugin cozy-client" :)

Crash--:
> Tu parles de quoi en particulier ?

De rien en fait, je pensais qu'on avait fait le travail pour AF après cette PR, mais en fait non, elle en découle.

nono:
@ptbrowne est-ce que tu veux rebaser cette PR?

ptbrowne:
I am leaving this open as it seems important that the bar closes its realtime sockets after logout, I have not checked if the behavior is still happening, cc @Crash--

PR #550, opened 2019-05-08
docs: Add example bar

ptbrowne:
Add a mobile bar example.

This lets us see a real example of how the bar should be used and 
can be used to develop on the bar without a full blown app that often hides
the essence of the API.

`$ yarn example` to launch example


examples/index.jsx
```diff
@@ -0,0 +1,97 @@
...
+import { Route, hashHistory } from 'react-router'
+import { MobileRouter } from 'cozy-authentication'
+import 'cozy-ui/transpiled/react/stylesheet.css'
+import appIcon from './icon.png'
+
+const minilog = window.minilog
+minilog.enable()
+minilog.suggest.allow('cozy-realtime', 'debug')
+
+// TODO ui: AppLinker should not fail so loudly without lampa.startapp plugin
```
Crash--: AppLinker will soon remove this dep ;). UL everywhere 💪

examples/index.jsx
```diff
@@ -0,0 +1,97 @@
...
+    </Layout>
+  )
+})
+
+const appInfo = {
+  name: 'Example bar',
+  slug: 'example-bar',
+  softwareID: 'io.cozy.example',
+  redirectURI: 'http://localhost:1234/auth',
+  universalLinkDomain: 'cozyexample://'
```
Crash--: Can we have something more realistic for ul domain ? Like https://cozyexample.com ? ;) 

Also you should have a protocol props with 'cozyexample://'

ptbrowne: Done, well spotted, I mixed up the two :)

package.json
```diff
@@ -67,6 +71,9 @@
     "postcss-discard-empty": "4.0.1",
     "postcss-import": "12.0.1",
     "postcss-loader": "3.0.0",
+    "react": "^16.8.6",
+    "react-dom": "^16.8.6",
```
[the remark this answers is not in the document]
ptbrowne: This is needed because the example uses react and react-dom, I'm changing this to use `my-react` and `my-react-dom`.

`parcel` is a bundler, it does not have a dependency on `react` nor on `react-dom` ;).

examples/Procfile
```diff
@@ -0,0 +1,3 @@
+build-bar: yarn watch:mobile
+serve-bar: cd dist; python -m SimpleHTTPServer 8000
```
CPatchane: This doesn't work with Python3 since we don't use `SimpleHTTPServer` but `http.server` instead

package.json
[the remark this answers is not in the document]
ptbrowne: I tried to change in the example to `react` and `react-dom` but since dependencies use `import React from 'react'` in their code,  it does not work. IMHO, this is not a big problem if developers have duplicates in their `node_modules`. Do you think it is a problem if there is a duplicate ?

examples/Procfile
ptbrowne: I updated the Procfile with `python3 -m http.server`, thanks for the tip :)

Crash--:
@ptbrowne Can you rebase and merge this PR?

ptbrowne:
Thanks for the reminder @Crash-- !

Crash--:
🍾

PR #552, opened 2019-05-09
WIP: Create CozyClient for realtime when bar initialized with cozyURL/token

ptbrowne:
Do not review yet. Untested.

Since the realtime lib used by the cozy-bar needs CozyClient, it should get
one even if the bar is initialized the old-fashioned way with only cozyURL
and token.
 
See this comment: https://github.com/cozy/cozy-bar/pull/545#discussion_r281610043


src/lib/realtime.js
```diff
@@ -1,38 +1,53 @@
-import realtime from 'cozy-realtime'
+import CozyRealtime from 'cozy-realtime'
+import CozyClient from 'cozy-client'
```
CPatchane: CozyClient is based on React, so we can't not import it at the beginning if we want to get React from the application, we have to require it on init, see https://github.com/cozy/cozy-bar/blob/master/src/index.jsx#L43

CPatchane: All libraries using React/ReactDom can't be imported at the top of files like this,we have to do that if we don't want to use React from node_modules and give time to the bar to check if the application [gave it React through the dom](https://github.com/cozy/cozy-bar/blob/ecab3bd98ae578f0869f3468e29f9649b96aee64/config/aliases/globalReact.js)

edas: That should not be a problem, we talk about apps that do not use cozy-client yet anyways (otherwise they should have been able to init the bar with it)

ptbrowne: Since merely importing CozyClient and not any HOC, webpack should do its job and not import React, right ? I am going to test this to be sure, WDYT ?

CPatchane: It should but I agree to test it to be sure. But if something would change on CozyClient side about React usage, we won't be protected here

kosssi:
Je pense que l'on se marche dessus @ptbrowne avec https://github.com/cozy/cozy-bar/pull/545/commits/ef90f9db2333bff281190f1b10a7c7984106dd7c

src/lib/realtime.js
edas: Do you have news on this ?

edas:
> Je pense que l'on se marche dessus @ptbrowne avec [ef90f9d](https://github.com/cozy/cozy-bar/commit/ef90f9db2333bff281190f1b10a7c7984106dd7c)

Bien vu. Quitte à choisir je préfère ton approche de l'initialiser dans l'init() de la bar plutôt que de le faire dans l'appel au realtime.

ptbrowne:
Done in https://github.com/cozy/cozy-bar/pull/545

PR #562, opened 2019-05-15
Append query string in AppLinker href prop 🚑

gregorylegarec:
App URLs with query string  were not properly encoded.

Let AppLinker manage the href with query string.

### DIFF
--- a/src/components/Apps/AppItem.jsx
+++ b/src/components/Apps/AppItem.jsx
@@ -55,7 +55,11 @@ export class AppItem extends React.Component {
     const dataIcon = app.slug ? `icon-${app.slug}` : ''
 
     return (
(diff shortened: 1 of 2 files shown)


ptbrowne:
Can we have a test for this please :) ?

gregorylegarec:
@ptbrowne what would you test ?

Crash--:
Testing a case like that: 

foo.bar/test/toto?fallback=https://fallback.fr/toto&from=test

gregorylegarec:
@Crash-- To me this test should be in AppLinker test suite, as here we are just appending the querystring to a given url (and tests for that already exist).

ptbrowne:
> @ptbrowne what would you test (I updated the snapshots) ?

I think http://fake.fr could have a query string, would it make the problem that we saw earlier more obvious ?

PR #563, opened 2019-05-15
fix: Add protocol to default stack url

cedricmessiant:
When the cozy-bar instantiates its own cozy-client, it needs a valid `uri`. The protocol was missing. 

Note that if the app initializes the bar with a `cozyURL`, it has to provide an URL with protocol : https://github.com/cozy/cozy-bar/blob/master/src/index.jsx#L200

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -106,7 +106,9 @@ const getDefaultStackURL = isPublic => {
     }
     return ''
   }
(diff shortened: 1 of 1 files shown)


ptbrowne:
A test would 💯 but I understand that the structure of the code makes it hard to do. LGTM.

PR #564, opened 2019-05-20
fix: Key should be uniq

Crash--:
URL should be enough is a vaste majority of use case. But in my case, I've several `Drive` installed and I got a few warnings complaining about the key not being unique.


Crash--:
[Console polluée dans drive](https://trello.com/c/T7jaiLPj/1938-console-pollu%C3%A9e-dans-drive)

PR #565, opened 2019-05-20
chore: Remove log for React dev tools message

Crash--:
React-dev-tools sends message but the bar log every message. 

See https://github.com/facebook/react-devtools/issues/812


Crash--:
[Console polluée dans drive](https://trello.com/c/T7jaiLPj/1938-console-pollu%C3%A9e-dans-drive)

PR #568, opened 2019-05-22
fix: cozy-client should be a peer dep

cedricmessiant:
We have issues with the cozy-stack-client version on drive. It seems to come from here

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     "babel-runtime": "6.26.0",
     "bundlesize": "0.17.1",
     "cozy-authentication": "^1.13.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -114,6 +113,9 @@
     "redux-persist": "5.10.0",
     "redux-thunk": "2.3.0"
   },
+  "peerDependencies": {
+    "cozy-client": "6.31.1"
+  },
```
Crash--: You should add it to dev-dep also

Crash--: And not sur about the fix on the version. `*` should be good, right?

PR #571, opened 2019-05-29
fix: spinners on buttons

GoOz:
Fix spinners on the cozy-bar by using `Button` component and removing conflicting styles.


GoOz:
[🔵 🍊 [HAR] Double spinner bouton connexion connecteur](https://trello.com/c/piN8FBTD/1425-%F0%9F%94%B5-%F0%9F%8D%8A-har-double-spinner-bouton-connexion-connecteur)

PR #575, opened 2019-05-31
Revert "fix(deps): update dependency @babel/core to v7.4.5"

CPatchane:
Reverts cozy/cozy-bar#491


Crash--:
Can you say why?

CPatchane:
I tried to merge a too old renovate with passed builds to get a new version but it broke the master build then. So I revert it since the build was broken, no broken version of the bar was published.

PR #580, opened 2019-06-04
WIP feat: Light bar

ptbrowne:
I am exploring what should we do to 

- diminish the size of the bar
- prevent integration surprise due to CSS order

Here are some leads:

### 1. Bar can use the same CSS stylesheet as the app

For this Cozy UI should not be transpiled with hashes. This way
the UI classes used in the bar should be the same as the app.

#### Benefits

  * Less integration surprise since there is only one stylesheet.
  * Since there is only one stylesheet, the overall JS/CSS served is less

#### Risks

  * The bar cannot use newer features from cozy-ui since it needs to rely
    on features that are available on all apps. In practice I think it will
    force us to keep cozy-ui up-to-date in our apps. 

#### Steps

1. Cozy UI stylesheet should be transpiled without hashes : https://github.com/cozy/cozy-ui/pull/1029
2. Update UI in banks / bar
3. Remove automatic importing of cozy-ui stylesheet in the bar https://github.com/cozy/cozy-bar/pull/580/commits/9221b9631a3b1d30b33b14f2fb0b2d5e11d20992

### 2. Lighter bundle without React and React dom

We can use webpack external feature to tell webpack that some dependencies should be provided by the
app via window.

This bundle could be used by apps using React / ReactDOM through a special asset cozy-bar.light.js

https://github.com/cozy/cozy-bar/pull/580/commits/7307bfe09d480a3d4108c27543f094046c75abc8

### 3. Bar no longer served by the stack

In web and mobile, same process, the bar is a dependency that should be updated and is no longer
served by the stack.

In practice the bar has to be updated in the apps because the native apps have no stack so it is will
only streamline the process between web / mobile. In this vein I have made an attempt at removing
the multibuild (target = browser or mobile) for the bar. I think it is an unnecessary complication
since window.cordova can be used to see if we are on native.

https://github.com/cozy/cozy-bar/pull/580/commits/c5a695403d5b11c36097d7c34b9c77aa9e406a48

### DIFF
--- a/config/webpack.js
+++ b/config/webpack.js
@@ -18,6 +18,7 @@ module.exports = (env = {}) => {
   return merge(
     require('./webpack.config.base.js'),
     addAnalyzer ? require('./webpack.config.analyzer.js') : {},
+    env.light ? require('./webpack.target.light.js') : {},
     require('./webpack.config.jsx.js'),
     require(production ? './webpack.config.prod' : './webpack.config.dev'),
     cssConfig({ filename, mobile, production }),
--- a/config/webpack.js
+++ b/config/webpack.js
@@ -1,7 +1,6 @@
 'use strict'
 
 const path = require('path')
-const webpack = require('webpack')
 const merge = require('webpack-merge')
 const vars = require('./webpack.vars')
 
@@ -26,12 +25,7 @@ module.exports = (env = {}) => {
       output: {
         filename: filename('js', mobile ? 'mobile' : ''),
         path: path.resolve(__dirname, '../dist')
-      },
-      plugins: [
-        new webpack.DefinePlugin({
-          __TARGET__: JSON.stringify(mobile ? 'mobile' : 'browser')
-        })
-      ]
+      }
     }
   )
 }
--- a/config/webpack.target.light.js
+++ b/config/webpack.target.light.js
@@ -1,0 +1,10 @@
+module.exports = {
+  externals: {
+    'cozy-client': 'CozyClient',
+    'cozy-stack-client': 'CozyStackClient',
+    'cozy-ui': 'CozyUI',
+    lodash: '_',
+    'my-react-dom': 'ReactDOM',
+    'my-react': 'React'
+  }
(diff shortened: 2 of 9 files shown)


edas:
Le dernier point est quelque chose à décider aussi avec les équipes produit.

Sur mobile on peut se permettre d'avoir une barre un peu différente suivant les app, sur le web c'est beaucoup moins le cas. Si on embarque la barre en dépendance ça veut dire qu'à tout moment il faut être capable de redéployer toutes les app de manière synchrone avec la nouvelle barre (nos app, mais aussi celles des tiers)

drazik:
We talked about this on zoom with @edas and @ptbrowne. We will first focus on the first point of this PR first: make the bar use the host's cozy-ui stylesheet.

The plan is:

1. Use transpiled cozy-ui in apps that don't use it yet (cozy-home and cozy-store it seems)
2. Test https://github.com/cozy/cozy-ui/pull/1029 and https://github.com/cozy/cozy-bar/pull/580/commits/9221b9631a3b1d30b33b14f2fb0b2d5e11d20992 on apps and fix what needs to be fixed
3. Merge https://github.com/cozy/cozy-ui/pull/1029 and update cozy-ui in all apps
4. Merge https://github.com/cozy/cozy-bar/pull/580/commits/9221b9631a3b1d30b33b14f2fb0b2d5e11d20992 and update the bar everywhere

After that, we should not have issues between cozy-bar's cozy-ui and apps' cozy-ui anymore.

I will work on this asap.

drazik:
Update: after further discussions, we agreed that a better solution exists. So, we will keep cozy-ui in the cozy-bar, but it will not used the transpiled version. Instead, it will transpile it itself with a specific className prefix and keep only the styles it uses to minimize the bundle size.

The benefit of doing this is that the cozy-bar's cozy-ui will be in its own scope that can't clash with the app's cozy-ui. This way, we don't have to consider the cozy-ui version used by the apps in order to fix the stylesheets issue.

PR #582, opened 2019-06-10
fix: Avoid cozy-ui stylesheet conflict with the apps

drazik:
This PR goal is to fix the bugs we have when we use a cozy-bar with a cozy-ui version that is not the same as the host app cozy-ui version.

This problems exists because the cozy-bar uses the transpiled version of cozy-ui, thus we could have two stylesheets that could override each other.

The solution explored here is to stop using the transpiled version of cozy-ui. The bar transpiles its own cozy-ui and use a custom classname scheme for cozy-ui's classnames. This way, there is no more conflicts between the bar's cozy-ui stylesheets and app's cozy-ui stylesheet.

This was tested on Banks, in which we had this issue: they are all gone.

### DIFF
--- a/config/webpack.config.base.js
+++ b/config/webpack.config.base.js
@@ -14,7 +14,7 @@ module.exports = {
     umdNamedDefine: true
   },
   resolve: {
-    extensions: ['.js', '.json', '.yaml'],
+    extensions: ['.js', '.json', '.yaml', '.jsx'],
     modules: [SRC_DIR, 'node_modules'],
     alias: {
       react: path.resolve(__dirname, 'aliases/globalReact'),
@@ -25,8 +25,11 @@ module.exports = {
   module: {
     rules: [
       {
-        test: /\.js$/,
-        exclude: /node_modules/,
+        test: /\.jsx?$/,
+        include: [
+          path.resolve(__dirname, '../src'),
+          path.dirname(require.resolve('cozy-ui'))
+        ],
         loader: 'babel-loader'
       },
       {
--- a/config/webpack.config.base.js
+++ b/config/webpack.config.base.js
@@ -18,8 +18,7 @@ module.exports = {
     modules: [SRC_DIR, 'node_modules'],
     alias: {
       react: path.resolve(__dirname, 'aliases/globalReact'),
-      'react-dom': path.resolve(__dirname, 'aliases/globalReactDOM'),
-      'cozy-ui/react': 'cozy-ui/transpiled/react'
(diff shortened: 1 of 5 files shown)


config/webpack.config.base.js
```diff
@@ -14,20 +14,22 @@ module.exports = {
     umdNamedDefine: true
   },
   resolve: {
-    extensions: ['.js', '.json', '.yaml'],
+    extensions: ['.js', '.json', '.yaml', '.jsx'],
     modules: [SRC_DIR, 'node_modules'],
     alias: {
       react: path.resolve(__dirname, 'aliases/globalReact'),
-      'react-dom': path.resolve(__dirname, 'aliases/globalReactDOM'),
-      'cozy-ui/react': 'cozy-ui/transpiled/react'
```
ptbrowne: If I am not mistaken, the most important part of this commit is the removal of the alias.

Nitpick: The name of the commit is a bit confusing, we are not "transpiling", we are bundling "cozy-ui" through its non transpiled version :)

drazik: I splitted the commit into two.

Crash--:
Related issues : 
#472 
#436

Seems a solid way to fix the issue. Let me build drive with it as we previously had some issues with the non transpiled version. 

Also, even if it's an internal changes we should considered this modification as at least a major version of the lib or as a breaking change.

drazik:
> Also, even if it's an internal changes we should considered this modification as at least a major version of the lib or as a breaking change.

I think even if I did small chore commits, we should squash it as a fix with breaking changes when merging it.

Crash--:
Seems to be ok for Drive/Android 👍

drazik:
> Do you know if the css from the cozy-bar contains any global classes from cozy-ui? Presumably it still has some utility classes?

Yes, the cozy-bar stylesheet contains utility classes that are used in cozy-ui components that are used in the cozy-bar.

y-lohse:
> Yes, the cozy-bar stylesheet contains utility classes that are used in cozy-ui components that are used in the cozy-bar.

Ok cool :) they shouldn't be a problem, but possibly the next issue to address.

config/webpack.config.base.js
```diff
@@ -25,8 +25,11 @@ module.exports = {
   module: {
     rules: [
       {
-        test: /\.js$/,
-        exclude: /node_modules/,
+        test: /\.jsx?$/,
+        include: [
+          path.resolve(__dirname, '../src'),
```
gregorylegarec: I think we must also add our overrides directory here. Probably not in this file however.

PR #584, opened 2019-06-12
Go back to config excluding node_modules

gregorylegarec:
The goal of this PR is to go back using an `exclude` clause. It will allow custom repositories to transpile override directories.

### DIFF
--- a/config/webpack.config.base.js
+++ b/config/webpack.config.base.js
@@ -26,10 +26,7 @@ module.exports = {
     rules: [
       {
         test: /\.jsx?$/,
-        include: [
-          path.resolve(__dirname, '../src'),
-          path.dirname(require.resolve('cozy-ui'))
-        ],
+        exclude: /node_modules\/(?!(cozy-ui\/react))/,
         loader: 'babel-loader'
       },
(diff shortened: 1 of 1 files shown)


config/webpack.config.base.js
```diff
@@ -26,10 +26,7 @@ module.exports = {
     rules: [
       {
         test: /\.jsx?$/,
-        include: [
-          path.resolve(__dirname, '../src'),
-          path.dirname(require.resolve('cozy-ui'))
-        ],
+        exclude: /node_modules\/(!?(cozy-ui))/,
```
CPatchane: Should we be more precise with a `cozy-ui/react` here?

gregorylegarec: I guess we should transpile everything in cozy-ui, not only the react folder.

ptbrowne: Why ? cozy-ui/react/transpiled is already transpiled.

CPatchane: @ptbrowne you mean `cozy-ui/transpiled/react` I think?

PR #602, opened 2019-06-28
Import utils CSS

gregorylegarec:
The goal is to provide Utils CSS to all apps


gregorylegarec:
I was not up-to-date about this point, thanks.

PR #608, opened 2019-07-03
Bundle cozy-ui

drazik:
This is a new attempt to transpile and bundle cozy-ui to avoid messing with the app's cozy-ui stylesheet.

Previous attempt (https://github.com/cozy/cozy-bar/pull/582) was done with a bad timing and bugs were found in the middle of a rush, so we had no choice but revert it.

### DIFF
--- a/config/webpack.config.base.js
+++ b/config/webpack.config.base.js
@@ -14,20 +14,19 @@ module.exports = {
     umdNamedDefine: true
   },
   resolve: {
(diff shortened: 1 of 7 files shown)


drazik:
> Did you tried to use it in other projects?

I tried it in banks and AF only. Thanks for the heads up, I will test it in other projects.

drazik:
I tested it in contacts and drive (web and android) : it works.

PR #613, opened 2019-07-16
fix: Avoid FOUC on mobile

drazik:
When sourcemap is enabled in css-loader, the style-loader injects a
`link` tag pointing to a blob generated dynamically. This leads to a
flash of unstyled content (FOUC) on the bar. Disabling the sourcemap
makes the style-loader inject a `style` tag and we don't have a FOUC
anymore.

### DIFF
--- a/config/webpack.config.inline-styles.js
+++ b/config/webpack.config.inline-styles.js
@@ -12,12 +12,11 @@ module.exports = ({ production }) => ({
         test: /\.styl$/,
         use: [
           {
-            loader: 'style-loader',
-            options: { sourceMap: true }
+            loader: 'style-loader'
           },
           {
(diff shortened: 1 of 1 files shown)


Crash--:
Should we check if we are in dev or prod mode and enable source map when we develop?

Crash--:
(also can you rebase? I just merged a fix for bundlesize)

drazik:
Actually that would be better to do that. But since https://github.com/cozy/cozy-banks/pull/1401 we use the unminified dev version of the bar in banks. So adding source map on the dev version would still produce a FOUC in banks mobile app :(

ptbrowne:
Nice catch !

PR #629, opened 2019-08-06
feat: Logs appearing every time and not really helpful

ptbrowne:
Those logs do not provide a way to fix the problem and are appearing every
time.


edas:
Could we leave a comment inside the code instead? I don't know this part of code but if Cedric did think it's an error, it's probably better to leave some comment for the future when someone will look into it

ptbrowne:
Done.

PR #664, opened 2019-11-29
feat: Check cozyclient version to call instance info

Crash--:
check cozyclient version before calling new functions 

Wanted to add a few tests for the component, but it requires more work since enzyme et enzyme react adapter are pretty old. An upgrade breaks other tests...

### DIFF
--- a/src/components/Settings/helper.js
+++ b/src/components/Settings/helper.js
@@ -1,3 +1,12 @@
-export const isFetching = requests => {
+import CozyClient from 'cozy-client'
+import compare from 'semver-compare'
+
+export const isFetchingQueries = requests => {
   return requests.some(request => request.fetchStatus === 'loading')
+}
+
+export const cozyClientCanCheckPremium = () => {
+  if (!CozyClient.version) return false
+  const result = compare(CozyClient.version, '8.3.0')
+  return result >= 0
 }
--- a/src/components/Settings/helper.js
+++ b/src/components/Settings/helper.js
@@ -5,8 +5,10 @@ export const isFetchingQueries = requests => {
   return requests.some(request => request.fetchStatus === 'loading')
 }
 
-export const cozyClientCanCheckPremium = () => {
-  if (!CozyClient.version) return false
-  const result = compare(CozyClient.version, '8.3.0')
+export const cozyClientCanCheckPremium = (forcedCozyClient = null) => {
+  const cozyClientToUse =
+    forcedCozyClient !== null ? forcedCozyClient : CozyClient
+  if (!cozyClientToUse.version) return false
+  const result = compare(cozyClientToUse.version, '8.3.0')
   return result >= 0
 }
--- a/src/components/Settings/index.jsx
+++ b/src/components/Settings/index.jsx
@@ -73,7 +73,7 @@ export class Settings extends Component {
       isFetching
     } = this.props
 
-    let shoulDisplayViewOfferButton = false
+    let shouldDisplayViewOfferButton = false
     let managerUrlPremiumLink
     let isFetchingFromQueries
     const canCheckPremium = cozyClientCanCheckPremium()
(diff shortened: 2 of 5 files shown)


test/components/Settings/helper.spec.js
```diff
@@ -18,6 +21,10 @@ describe('Settings Helper', () => {
     const fakeRequest2 = {
       fetchStatus: 'loaded'
     }
-    expect(isFetching([fakeRequest1, fakeRequest2])).toBe(false)
+    expect(isFetchingQueries([fakeRequest1, fakeRequest2])).toBe(false)
+  })
+
+  it('should return true for cozyClientCanCheckPremium', () => {
```
ptbrowne: I think this test should be clearer if cozyClientCanCheckPremium would take CozyClient in argument. This way you could test both true/false behaviors.

yarn.lock
```diff
@@ -4428,13 +4428,13 @@ cozy-authentication@1.19.1:
     snarkdown "1.2.2"
     url-polyfill "1.1.7"
 
-cozy-client@8.1.0:
-  version "8.1.0"
-  resolved "https://registry.yarnpkg.com/cozy-client/-/cozy-client-8.1.0.tgz#3dbaa6e6454241f186e8b242990129f4a22d3334"
-  integrity sha512-Xy4cpxTIdu1jVUetgd6aG+n4qgYKqhXTkxnSq/OPt+Px8kS1XH3aQMqPT/bBb81sM5RGz1qtkPP8H9SHBbNo6g==
+cozy-client@^8.3.1:
```
ptbrowne: Is it expected not to have pinned the version ?

src/components/Settings/index.jsx
```diff
@@ -66,29 +69,40 @@ export class Settings extends Component {
...
+    let isAllFetchingAreDone = false
+    if (!canCheckPremium) {
+      isAllFetchingAreDone = !isFetching
+    } else {
+      isAllFetchingAreDone = !isFetchingFromQueries && !isFetching
+    }
+
     const { opened } = this.state
-    const openMenu = opened && !isCurrentlyFetching
+    const openMenu = opened && isAllFetchingAreDone
```
ptbrowne: isAllFetchingDone ou areAllFetchingDone but not both ;)

src/components/Settings/index.jsx
```diff
@@ -66,29 +69,40 @@ export class Settings extends Component {
...
+        instanceQuery,
+        contextQuery
+      ])
+      if (!isFetchingFromQueries) {
+        const data = {
+          context: contextQuery,
+          diskUsage: diskUsageQuery,
+          instance: instanceQuery
+        }
+        shoulDisplayViewOfferButton = instanceModel.shouldDisplayOffers(data)
```
ptbrowne: s/shoulDisplayViewOfferButton/shouldDisplayViewOfferButton

src/components/Settings/helper.js
```diff
@@ -1,3 +1,12 @@
-export const isFetching = requests => {
+import CozyClient from 'cozy-client'
+import compare from 'semver-compare'
+
+export const isFetchingQueries = requests => {
```
ptbrowne: Nit: I think it would be better / more flexible/composable to extract the

request.fetchStatus === 'loading' into isCollectionLoading 

and then to call directly `requests.some(isCollectionLoading)`

I think it is clearer when reading.

Crash--: Nice idea, I think that this kind of utilities functions should be served by cozyclient itself, wdyt?

src/components/Settings/index.jsx
Crash--: same again and again 🤦‍♂

test/components/Settings/helper.spec.js
Crash--: Thanks :)

yarn.lock
Crash--: Hmm nop. So using package@latest do no pin the version. Interesting.

src/components/Settings/helper.js
ptbrowne: Yes, it's been a long time that we have isCollectionLoading in Banks, PR incoming.

Crash--: Thanks 😊

ptbrowne: https://github.com/cozy/cozy-client/pull/576

PR #672, opened 2019-12-02
fix: Better support for old cozyclient version

Crash--:
On old CozyClient version, we can't use queryConnect since we have errors with ObersableQueries. 

In order to avoid that issue, we export the "new component" only if we can checkPremium. (tested on current contact stable) 

Also, we can't read cozyclient's version from the static import of cozy-client. We need to read it from the cozy-client's instance passed to the bar

### DIFF
--- a/src/components/Settings/helper.js
+++ b/src/components/Settings/helper.js
@@ -1,13 +1,25 @@
-import CozyClient from 'cozy-client'
 import compare from 'semver-compare'
-
+import client from 'lib/stack-client'
 export const isFetchingQueries = requests => {
   return requests.some(request => request.fetchStatus === 'loading')
 }
-
+/**
+ *
+ * @param {cozyClient} forcedCozyClient only used to test purpose
+ *
+ * We can not read `version` from `import CozyClient from cozy-client`
+ * since in that case, we'll read version from the cozy-bar node modules
+ * and not from the app one.
+ *
+ * In order to avoid this issue, we get the cozyclient instance from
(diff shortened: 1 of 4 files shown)


src/components/Settings/helper.js
```diff
@@ -1,13 +1,25 @@
...
+ * In order to avoid this issue, we get the cozyclient instance from
+ * lib/stack-client (cozyclient's instance passed by the app to the bar),
+ * then read the constructor and then read the version from it
+ */
 export const cozyClientCanCheckPremium = (forcedCozyClient = null) => {
+  const usedClient = client.getClient() ? client.getClient().constructor : {}
   const cozyClientToUse =
-    forcedCozyClient !== null ? forcedCozyClient : CozyClient
+    forcedCozyClient !== null ? forcedCozyClient : usedClient
+
```
y-lohse: Nit; but naming the vars `usedClientClass` or something like that might be good.

src/components/Settings/index.jsx
```diff
@@ -5,7 +5,11 @@ import { translate } from 'cozy-ui/react/I18n'
 import { Button } from 'cozy-ui/react/Button'
 import { queryConnect } from 'cozy-client/dist'
 import { models } from 'cozy-client'
-const { instance: instanceModel } = models
+let instanceModel = undefined
+if (models) {
+  instanceModel = models.instance
+}
+
```
y-lohse: Works but I still like `get(models.instance)` better

PR #673, opened 2019-12-03
Better app names

y-lohse:
This PR brings the app list in line with the behavior on the home page:

- No Cozy prefix
- Support translated names
- Display Settings

I did 475512012cd1cb831ad2f2df425b13999114dcff anticipating I'd need it for using `applicationsModel`but didn't end up using it. I still think it's better like this but we can also remove this commit from the PR.

### DIFF
--- a/src/components/Apps/AppItem.jsx
+++ b/src/components/Apps/AppItem.jsx
@@ -1,12 +1,19 @@
 import React from 'react'
+import get from 'lodash/get'
 import { appShape } from 'proptypes/index'
+import { models } from 'cozy-client'
 
 import AppIcon from 'cozy-ui/react/AppIcon'
 import AppLinker from 'cozy-ui/react/AppLinker'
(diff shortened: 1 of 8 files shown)


src/components/Apps/AppItem.jsx
```diff
@@ -57,9 +58,19 @@ export class AppItem extends React.Component {
...
     const dataIcon = app.slug ? `icon-${app.slug}` : ''
+    const getAppDisplayName = get(
+      models,
+      'applications.getAppDisplayName',
+      app => {
+        return app.namePrefix && app.namePrefix.toLowerCase() !== 'cozy'
+          ? `${app.namePrefix} ${app.name}`
+          : app.name
+      }
+    )
```
Crash--: This declaration can be done outside of the render() method. It'll avoid the  `get` cost at every render

PR #674, opened 2019-12-06
feat: Display My offer if needed

Crash--:
the idea: 
- If we haven't an offer : We display "View offers"
- If we already have an offer, then we display "My offer"

### DIFF
--- a/src/components/Settings/index.jsx
+++ b/src/components/Settings/index.jsx
@@ -1,13 +1,32 @@
 import React, { Component } from 'react'
 import { connect } from 'react-redux'
 import { compose } from 'redux'
+import get from 'lodash/get'
+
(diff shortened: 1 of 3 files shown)


src/components/Settings/index.jsx
```diff
@@ -93,8 +113,14 @@ export class Settings extends Component {
           diskUsage: diskUsageQuery,
           instance: instanceQuery
         }
+        shouldDisplayViewOfferButton =
+          instanceModel.shouldDisplayOffers(data) || hasAnOffer(data)
```
y-lohse: Are there still cases where we *don't* display the button?

Crash--: Yep. If cozy-client version is a bit late

PR #682, opened 2020-03-31
feat: Better default app support for self hosted cozy

Crash--:
We make a request to /settings/context to get the "Default App", in order to make the link "Back to home" in the cozy bar. But with a self hosted cozy, we don't have a context by default and we don't have a default home slug defined.

So two solutions: 
- self hosted can create a context and add a default_redirection param equal to "home/"
- Edit the cozy bar to get the "HOME" slug if the default_redirection is not defined. 

I renamed homeApp to DefaultApp since we can have a different app

### DIFF
--- a/src/lib/reducers/apps.js
+++ b/src/lib/reducers/apps.js
@@ -49,7 +49,7 @@ export const fetchApps = () => async dispatch => {
     if (!rawAppList.length)
       throw new Error('No installed apps found by the bar')
     // TODO load only one time icons
-    await dispatch(setHomeApp(apps))
+    await dispatch(setDefaultApp(apps))
     await dispatch(receiveAppList(apps))
   } catch (e) {
     dispatch({ type: FETCH_APPS_FAILURE })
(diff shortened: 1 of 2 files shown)


src/lib/reducers/apps.js
```diff
@@ -58,19 +58,37 @@ export const fetchApps = () => async dispatch => {
...
+        const HOME_APP_SLUG = 'home'
+        homeApp = appsList.find(app => app.slug === HOME_APP_SLUG)
+      } else {
+        const slugRegexp = /^([^/]+)\/.*/
+        const defaultAppSlug =
+          defaultRedirection &&
+          defaultRedirection.match(slugRegexp) &&
+          defaultRedirection.match(slugRegexp)[1]
+        homeApp = appsList.find(app => app.slug === defaultAppSlug)
+      }
```
ptbrowne, line 86: Nit: This could be extracted. It would be more easily testable.

src/lib/reducers/apps.js
```diff
@@ -58,19 +58,37 @@ export const fetchApps = () => async dispatch => {
...
+      //so let's use hardcoded home slug if needed
+      if (!defaultRedirection) {
+        const HOME_APP_SLUG = 'home'
+        homeApp = appsList.find(app => app.slug === HOME_APP_SLUG)
+      } else {
+        const slugRegexp = /^([^/]+)\/.*/
+        const defaultAppSlug =
+          defaultRedirection &&
+          defaultRedirection.match(slugRegexp) &&
+          defaultRedirection.match(slugRegexp)[1]
```
ptbrowne, line 84: Nit: Regex match is done twice, not very important here though.

src/lib/reducers/apps.js
```diff
@@ -58,19 +58,37 @@ export const fetchApps = () => async dispatch => {
...
+        homeApp = appsList.find(app => app.slug === defaultAppSlug)
+      }
+
+      if (!homeApp) {
+        return appsList
+      }
```
ptbrowne, line 90: Nit: I am a bit uncertain here what's the effect of not having the final dispatch is. It seems a bit awkward to return an array here where we return the result of the dispatch otherwise. Are we sure we need to return something ? The function is called `set` so AMHA it should only have a side effect and not return anything.

Crash--:
At first it was just a quick fix for a self hosted, but you're right concerning the test and a better code, it doesn't cost so much

src/lib/reducers/apps.spec.js
```diff
@@ -0,0 +1,59 @@
...
+import stack from 'lib/stack.js'
+import { setDefaultApp } from './apps'
+
+const cozyURL = 'https://test.mycozy.cloud'
+const token = 'mytoken'
+const fakeStackClient = {
+  uri: cozyURL,
+  token: { token }
+}
+describe('app reduceur', () => {
```
ptbrowne, line 12:
```suggestion
describe('app reducer', () => {
```

PR #696, opened 2020-07-17
Remove distinction between browser and mobile builds

ptbrowne:
Draft: I've remarked this issue https://github.com/cozy/cozy-bar/pull/582/files that I think is important to take into account.

- Remove distinction between browser and mobile builds

Only 1 build is used. isMobileApp is used to differentiate between
mobile and browser platforms at runtime. More info in the commit.

This is a breaking change as cozy-bar.mobile.js is no longer shipped.

- Use flag to reduce activate redux logging of the bar

By default, the bar would attach a logger to redux. This pollutes the
app consoles that are seldom interested in what the bar is doing.
Here, with the flag, we can activate/deactivate logging easily.


Crash--:
Est-ce que le fait d'avoir un seul bundle a changé la taille du bundle web par exemple ?

ptbrowne:
Non, on avait pas de différence entre le code web et mobile puisqu'on utilise des checks au runtime depuis un moment.

nono:
@ptbrowne do you know why the travis build is red?

ptbrowne:
No, I do not plan to work on this further during the next 2 weeks.

Crash--:
Is it still relevant @ptbrowne ?

ptbrowne:
No :)

PR #697, opened 2020-07-22
chore: Move parcel as dev dep

ptbrowne:
Trying to solve problems with babel imports inside cozy-contacts
adfter upgrade to cozy-scripts@4, I remarked that parcel was installed
in cozy-contacts due to the cozy-bar declaring it as a normal
dependency.


ptbrowne:
I've merging since the change is pretty scoped, and since parcel is not real dependency.

Crash--:
note: This is not a `chore` commit ^^. We're changing the code we're releasing since parcel is not present anymore.

ptbrowne:
Yes, I've seen that after it did not trigger a release :(

PR #702, opened 2020-08-14
fix: Do not use module.exports and imports in the same file

ptbrowne:
Otherwise we have an errors "exports" is read-only


JF-Cozy:
The build seems to be broken on this one :(

PR #703, opened 2020-08-17
chore: Remove travis-deploy-once

ptbrowne:
travis-deploy-once causes a [deployment problem](https://travis-ci.org/github/cozy/cozy-bar/builds/718566049#L1113)
since it depends on a @babel/preset-env. This causes a dependency
resolving problem since the @babel/preset-env of travis-deploy-once
cannot find the @babel/plugin-transform-regenerator that is under
the node_modules of the root @babel/preset-env.

It's a bit clearer when seeing the structure:

```
node_modules
    travis-deploy-once
        node_modules
            @babel/preset-env
                ❌ cannot find the @babel/plugin-transform-regenerator since
                    it needs to go up AND down, which is not covered by
                    the node dependency solving mechanism
    @babel/preset-env
        node_modules
            @babel/plugin-transform-regenerator
```

Anyway, travis-deploy-once is deprecated and [semantic-release advocates
to use build stages](https://github.com/semantic-release/travis-deploy-once).

We do no need build stages here since we are not testing on multiple
platforms.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ before_deploy:
 deploy:
   provider: script
   skip_cleanup: true
-  script: npm run travis-deploy-once "npm run semantic-release"
+  script: npm run semantic-release
   on:
     branch: master
     repo: cozy/cozy-bar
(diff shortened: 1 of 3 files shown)


JF-Cozy (approved):
LGTM 👍

PR #706, opened 2020-10-16
feat: Ability to show something in the search part of the bar

ptbrowne:
With this, we will be able to show a custom search bar in the bar.

Should Drive use this capacity to show its search bar (this way the bar would not carry the code that is at the moment very specific to drive) ?

### DIFF
--- a/test/lib/__snapshots__/api.spec.jsx.snap
+++ b/test/lib/__snapshots__/api.spec.jsx.snap
@@ -25,6 +25,12 @@ exports[`api spec should set content of the store via declarative api 3`] = `
 </div>
 `;
 
+exports[`api spec should set content of the store via declarative api 4`] = `
+<div>
+  Search
+</div>
+`;
+
 exports[`api spec should set content of the store via imperative api 1`] = `
 <span
   dangerouslySetInnerHTML={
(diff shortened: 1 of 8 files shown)


test/lib/__snapshots__/api.spec.jsx.snap
```diff
@@ -25,6 +25,8 @@ exports[`api spec should set content of the store via declarative api 3`] = `
 </div>
 `;
 
+exports[`api spec should set content of the store via declarative api 4`] = `<div />`;
```
Crash--, line 28: what does that test ? ^^

Crash-- (approved):

Crash--:
> Should Drive use this capacity to show its search bar (this way the bar would not carry the code that is at the moment very specific to drive) ?

At first, we thought to be able to provide a generic way to search in all your cozy. I never read the code, just took a look to write this comment but it seems not specific to drive even if it's only used in Drive ATM. 

Does this removal will have a "big" impact in the bundle size? 

Let's discuss that point next week

test/lib/__snapshots__/api.spec.jsx.snap
ptbrowne: I had forgotten to put a content for the search. Fixed in https://github.com/cozy/cozy-bar/pull/706/commits/789de16c4f2c8064a9a79562945252a6a9a7571d

ptbrowne:
I've added the point to the card in Trello.

PR #708, opened 2020-10-29
Only 1 build for web and mobile

ptbrowne:
BREAKING CHANGE : no more mobile build

Since the code was mostly the same, removing the mobile specific
build makes it easier to reason about.

This PR also includes improvements that will facilitate the shipping
of the bar as a library / smaller build (background images).


src/components/Apps/index.jsx
```diff
@@ -50,9 +49,7 @@ class Apps extends Component {
     const { opened } = this.state
     return (
       <nav
-        className={`coz-nav coz-nav-apps${
-          replaceTitleOnMobile ? ' coz-bar-hide-sm' : ''
-        }`}
+        className='coz-nav coz-nav-apps'
```
Crash--, line 52: Is it a breaking change? It seems that in Drive / Photo we use this `replaceTitleOnMobile` props 🤔

src/lib/store/index.js
```diff
@@ -16,26 +16,15 @@ const config = {
...
-  const parts = ['[COZY-BAR]']
-
-  parts.push(`${String(action.type)}`)
-  if (time) parts.push(`@ ${time}`)
-
-  return parts.join(' ')
-}
-
 const loggerMiddleware = createLogger({
-  titleFormatter: barTitleFormatter,
```
Crash--, line 30: I though you added support for group/end https://github.com/cozy/minilog/commit/80414d3d28b609c7452c937524c1838928dc6a06

Crash-- (approved):

src/components/Apps/index.jsx
ptbrowne: There was no difference in behavior if I removed or kept the prop.

src/lib/store/index.js
ptbrowne: Ah yes, but group and groupEnd are always shown, without consideration for the current log level (since the group call does not know about the message level), which pollutes the console.

PR #709, opened 2020-11-02
Transpilation

ptbrowne:
Transpilation will make it possible to use a non bundled version
of the cozy-bar in an app, for example on mobile environments.

It also makes it faster to develop on the bar since you only have to transpile files instead of bundling them.


Crash-- (approved):
👏 looks solid

PR #729, opened 2021-07-28
docs: Add some info to use the bar with Context

JF-Cozy:
see commit message

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,26 +3,21 @@
 [![NPM Licence shield](https://img.shields.io/npm/l/cozy-bar.svg)](https://github.com/cozy/cozy-bar/blob/master/LICENSE)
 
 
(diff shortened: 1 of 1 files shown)


flohhhh (approved):

README.md
```diff
@@ -97,7 +94,7 @@ const MyConnectedComponent = connect(mapStateToProps, mapDispatchToProps, MyComp
 </BarLeft>
 ```
 
-Instead, you can do something like this:
+Instead, you can do something like this (for Context you can pass the values as prop):
```
Crash--, line 97: Or you can use https://github.com/cozy/cozy-ui/blob/master/react/BarContextProvider/index.jsx ? ;) ( cc @JF-Cozy )

PR #730, opened 2021-09-03
Fix/icons

Crash--:
See commit's message.

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,7 +99,7 @@
     "cozy-flags": "2.4.1",
     "cozy-interapp": "0.4.9",
     "cozy-realtime": "3.2.1",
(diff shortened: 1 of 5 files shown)


acezard (approved):

package.json
```diff
@@ -99,7 +99,7 @@
     "cozy-flags": "2.4.1",
     "cozy-interapp": "0.4.9",
     "cozy-realtime": "3.2.1",
-    "cozy-ui": "40.1.0",
+    "cozy-ui": "^51.8.0",
```
acezard, line 102: Aren't we using fixed versions for cozy libs?

Crash--: Let's discuss that on Tuesday but no ;)

PR #739, opened 2022-02-11
test: increase expected bundlesize

trollepierre:


trollepierre:
this PR will be draft while waiting for more explanation about de +60kb

Crash--:
done in #742  but we still need to understand why

PR #746, opened 2022-02-25
feat: Upgrade cozy-client

acezard:
To match cozy-banks


trollepierre (approved):
hoping test won't fail :-)

PR #748, opened 2022-02-28
feat: Allow cozy-flag to display home button

acezard:
Adds convenience for debugging purposes

### DIFF
--- a/src/components/Apps/ButtonCozyHome.jsx
+++ b/src/components/Apps/ButtonCozyHome.jsx
@@ -1,15 +1,17 @@
 import React from 'react'
 
+import flag from 'cozy-flags'
+
 import IconCozyHome from './IconCozyHome'
 
 export const ButtonCozyHome = ({ webviewContext, homeHref }) => {
-  if (webviewContext)
+  if (webviewContext || flag('flagship.debug'))
     return (
       <a
         onClick={() => {
           webviewContext.call('backToHome')
         }}
-        className="coz-nav-apps-btns-home"
+        className="coz-nav-apps-btns-home --is-flagship"
       >
         <IconCozyHome className="coz-nav-apps-btns-home-svg" />
       </a>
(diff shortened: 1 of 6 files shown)


src/lib/logger.js
```diff
@@ -1,4 +1,9 @@
+import set from 'lodash.set'
```
trollepierre, line 1: should we add a library, just for a small set ?

can we instead just use a simple === ?

this would make build lighter.

src/lib/logger.js
```diff
@@ -1,4 +1,9 @@
...
+
 import minilog from '@cozy/minilog'
+import flag from 'cozy-flags'
+
+set(window, `cozy.debug.flagship`, () => flag('flagship.debug', true))
```
trollepierre, line 6: did you test what happen when cozy is empty? When cozy.debug is empty? when flagship is defined to a wrong value? 

could we document these behaviours with unit tests 🙏 ?

src/components/Apps/ButtonCozyHome.jsx
```diff
@@ -1,9 +1,11 @@
 import React from 'react'
 
+import flag from 'cozy-flags'
+
 import IconCozyHome from './IconCozyHome'
 
 export const ButtonCozyHome = ({ webviewContext, homeHref }) => {
-  if (webviewContext)
+  if (webviewContext || flag('flagship.debug'))
```
trollepierre, line 8: could you add a simple test on that feature? 🙏

src/lib/logger.js
acezard: well lodash will handle that, that's in part why I don't think tests are needed here (lodash is already tested)

src/lib/logger.js
acezard: well cozy-flags is used like that from what I saw, just wanted to stay in the same line

trollepierre: cozy-flags does not have lodash.set as dependencies :x:

PR #759, opened 2022-07-05
V7 stable

acezard:


Crash--:
What is this PR? Can you explain it?

acezard:
> What is this PR? Can you explain it?

github was spamming with creating a PR but I don't intend to merge it for now (or ever probably, not sure).
I'm basically creating a 'master-like' branch for v7, to replace the mess that was the v7 updating process existing before. This will be documented after the fact

PR #766, opened 2022-07-19
fix: Make sure searchbar doesn't render

acezard:
If searchbar ever renders in FlagshipApp, this will currently trigger
a breaking bug with iframes CSP issues.
As of now we just ask the cozybar to not render the searchbar when
we're in flagship so the issue does not exist


Crash--:
You can add that this is a workaround since the current condition should work.

PR #769, opened 2022-08-01
v7-stable to v8

acezard:
The goal is not to merge v7 into v8, but to keep track of the differences between the two versions, and ideally pushing/pulling what should be mutualised between the two versions


trollepierre:
reopen in https://github.com/cozy/cozy-bar/pull/772

PR #773, opened 2022-08-18
feat(SearchBar): fetch and open note once clicked on a suggestion

trollepierre:
Problem was detected by:
https://github.com/cozy/cozy-drive/pull/2663?notification_referrer_id=NT_kwDOAH-dRrI0MTM4NzMwMjUyOjgzNjMzMzQ&notifications_query=is%3Aunread#discussion_r946530786

notes URL were not used to redirect, once clicked.

By fixing that bug, I realized the postMessage onSelect function could not be sent through postMessage. Becasue postMessage serializes the object before sending it.
[An object could not be cloned](https://stackoverflow.com/questions/42376464/uncaught-domexception-failed-to-execute-postmessage-on-window-an-object-co)

Fetching URL is now done inside Cozy-Bar scope

### DIFF
--- a/src/components/SearchBar.jsx
+++ b/src/components/SearchBar.jsx
@@ -4,6 +4,7 @@ import Autosuggest from 'react-autosuggest'
 import debounce from 'lodash.debounce'
 import { fetchRawIntent } from 'lib/intents'
 import logger from 'lib/logger'
+import { models, withClient } from 'cozy-client'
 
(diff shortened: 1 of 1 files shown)


trollepierre:
@Crash-- : you have approuved #774 but this is the PR that adds the feature in v7-stable branch. Could you approuved it?

Crash-- (approved):

PR #774, opened 2022-08-18
 feat(SearchBar): fetch and open note once clicked on a suggestion

trollepierre:
See PR description of https://github.com/cozy/cozy-bar/pull/773

### DIFF
--- a/src/components/SearchBar.jsx
+++ b/src/components/SearchBar.jsx
@@ -4,6 +4,7 @@ import Autosuggest from 'react-autosuggest'
 import debounce from 'lodash.debounce'
 import { fetchRawIntent } from 'lib/intents'
 import logger from 'lib/logger'
+import { models, withClient } from 'cozy-client'
 
 const INTENT_VERB = 'OPEN'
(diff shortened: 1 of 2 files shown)


src/components/Bar.jsx
```diff
@@ -237,7 +237,10 @@ export class Bar extends Component {
           {barLeft || this.renderLeft()}
           {barCenter || this.renderCenter()}
           <div className="u-flex-grow">
-            {barSearch || (searchBarEnabled ? <SearchBar /> : null)}
+            {barSearch ||
+              (searchBarEnabled ? (
+                <SearchBar client={this.props.cozyClient} />
```
Crash--, line 242: What is the gain to pass client as props? Why not using `withClient` as the rest of our codebase?

trollepierre: I will use `useClient()` , it will be more straightforward.

trollepierre: fixed

Crash-- (approved):

PR #775, opened 2022-08-23
fix(warning): mark as UNSAFE component will mount

trollepierre:
it removes warnings in Drive cause by Searchbar


Crash-- (approved):
Approved, but can you change the commit's title? it's not a `chore` commit since we change production code. fix / feat as you want since it can be both, but not chore.

PR #778, opened 2022-09-02
fix(Search): withClient provoked Minified React Error

trollepierre:
you have to use client coming from prop because with hoc or hooks
it might cause this Minified React Error


Crash--:
> I need to create a PR for v8

I don't think we'll have the issue for v8, since the React is shared. 


Can you change the commit message please: "Prefer to pass props client to handle this error". 

This is not "prefer" its more "you have to use client coming from prop because with hoc or hooks..."

Crash-- (approved):

PR #781, opened 2022-09-29
feat: Launch stack request on focus

Crash--:
Since we send a (fake) message via postMessage,
Cozy Drive's intent will catch it and
send the _all_docs request. Like that
we warm up the query and the data and
we expect to give a better UX like that.

### DIFF
--- a/src/components/SearchBar.jsx
+++ b/src/components/SearchBar.jsx
@@ -173,6 +173,19 @@ class SearchBar extends Component {
   }
 
   changeFocusState = focused => {
+    // hack in order to launch the stack request on drive.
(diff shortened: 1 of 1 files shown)


paultranvan:
That's truly not beautiful, but still a clever trick :smile: 
If I understand correctly, the expected gain is to start the query as soon as the search has the focus, rather than waiting a user input? So in the worst case, we launch a query for nothing (user didn't type anything), and in the best case, the user has already all the data in RAM after typing. And I would assume we would be closer to the best case in general.

paultranvan (approved):

PR #790, opened 2022-11-28
feat: Update settings sections

zatteo:
To fit with the structure of the new settings app.

```
### ✨ Features

* Add and move links in dropdown settings
```


src/components/Settings/SettingsContent.jsx
```diff
@@ -14,38 +14,93 @@ const SettingsContent = ({
   onClaudy,
   isDrawer = false,
```
zatteo, line 15: What is the meaning of the isDrawer ? Is there a mobile responsive version ?

Crash--: Yep. But I don't know if this api is still used :/

src/components/Settings/SettingsContent.jsx
```diff
@@ -14,38 +14,93 @@ const SettingsContent = ({
...
+                </a>
+              </li>
+            </>
+          )}
+        </ul>
+        <ul className="coz-nav-group">
+          <li className="coz-nav-settings-item">
+            <a
+              role="menuitem"
+              href={`${settingsAppURL}#/permissions/slug`}
```
Crash--, line 72: can you hide this link behind this flag? https://github.com/cozy/cozy-settings/blob/e7ee451bc5feb4540f47a00e0683f247658338a4/src/components/Sidebar.tsx#L80

src/components/Settings/SettingsContent.jsx
zatteo: It seems to be used at the root of the Bar for the moment

Crash-- (approved):

PR #805, opened 2022-12-08
fix: Remove duplicate WebviewIntentProvider

zatteo:
```
### Tech
 * WebviewIntentProvider is already added in apps using the bar.
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -116,7 +116,6 @@
     "semver-compare": "^1.0.0"
   },
   "peerDependencies": {
-    "cozy-intent": ">=1.7.0",
     "@cozy/minilog": "^1.0.0",
     "cozy-client": "^34.1.5",
     "cozy-device-helper": "2.6.0",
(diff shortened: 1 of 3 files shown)


src/index.jsx
```diff
@@ -26,7 +26,6 @@ import {
 import enLocale from 'locales/en.json'
 import frLocale from 'locales/fr.json'
 import esLocale from 'locales/es.json'
-import { WebviewIntentProvider } from 'cozy-intent'
```
Crash--, line 29: Can you remove it from the package.json ?

package.json
```diff
@@ -53,7 +53,6 @@
     "babel-plugin-transform-define": "^2.0.0",
     "babel-preset-cozy-app": "^1.9.0",
     "cozy-authentication": "1.19.1",
-    "cozy-intent": "^1.7.0",
```
acezard, line 56: test/build will be broken if `cozy-ui@77.9.1` doesn't have access to `cozy-intent`

acezard (changes requested):

Crash-- (approved):

acezard (approved):

PR #809, opened 2023-01-03
feat: Add Bar Component

cballevre:
The idea behind this PR is to expose the bar component via `<BarComponent>` so it can be integrated in the react-dom. 
The expected benefits are to be able to use hook from the application with is imported in for example `useRouter`. This also make specific development of custom part more easy because it shares the same context.

For the moment, nothings works really but this a WIP PR to keep track of the exploration.


cballevre:
Old work for the new version of the cozy-bar. The final PR is https://github.com/cozy/cozy-bar/pull/833

PR #828, opened 2023-07-05
feat: Update cozy-ui and all its dependencies

zatteo:
BREAKING CHANGE:
- you need to upload cozy-ui >= 88.2.1
- you need to upload cozy-client >= 38.9.2

@Crash-- do we handle breaking changes the same way on cozy-bar repository ?


Crash--:
> @Crash-- do we handle breaking changes the same way on cozy-bar repository ?
for v8 and +, yes. 

for v7, no since we're bundling the cozy-bar to be autonomous.

Crash-- (approved):

PR #830, opened 2023-08-11
feat: Remove Claudy v7

cballevre:


src/queries/index.js
```diff
@@ -1,16 +1,16 @@
...
   query: () => Q('io.cozy.settings').getById('context'),
-  as: 'contextQuery'
+  as: 'io.cozy.settings/context'
 }
 
 export const diskUsageReq = {
   query: () => Q('io.cozy.settings').getById('disk-usage'),
-  as: 'diskUsageQuery'
+  as: 'io.cozy.settings/disk-usage'
 }
```
Crash--, line 16: If I'm not mistaken, the cozy-client used by the cozy-bar comes from the app itself. So to be able to use `getById('io.cozy.settings.instance')` the app should pass a "relatively new" cozy-client. I don't think we have this version on all our apps... 

I don't know what will happen if the version is not right? We need at least this version of cozy-client https://github.com/cozy/cozy-client/commit/3a35f4fd34a74b1133bb9fc36a1f915c69e4122f

cballevre: You're not mistaken, the cozy-client instance does come from the application. I hadn't thought about our old application when I cherry-picked the commit so I removed it from this PR

Crash-- (approved):

PR #838, opened 2024-03-18
feat: add translations for banner moved (VO-46)

acezard:
If the html document contains a "moved" descriptor, we will display the correct translation.
Also, we won't display banner CTA as there is no link available in the metadata.

### DIFF
--- a/src/locales/fr.json
+++ b/src/locales/fr.json
@@ -42,6 +42,9 @@
     "tos-updated": {
       "description": "Dans le cadre du RGPD, Cozy Cloud met à jour ses Conditions Générales d'Utilisation qui ont pris effet le 25 Mai 2018",
       "CTA": "Lire maintenant"
+    },
+    "moved": {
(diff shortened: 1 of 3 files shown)


src/locales/fr.json
```diff
@@ -42,6 +42,9 @@
     "tos-updated": {
       "description": "Dans le cadre du RGPD, Cozy Cloud met à jour ses Conditions Générales d'Utilisation qui ont pris effet le 25 Mai 2018",
       "CTA": "Lire maintenant"
+    },
+    "moved": {
+      "description": "Ce Cozy a été déplacé vers une nouvelle adress."
```
cballevre, line 47: Same as for the v7, there is a e missing for the french translation of address

cballevre (approved):

src/locales/fr.json
acezard: updated

PR #840, opened 2024-03-18
feat: update link for subscription flag (VO-386)

acezard:
Will redirect to store if settings.subscription flag is present

### DIFF
--- a/src/components/Settings/index.jsx
+++ b/src/components/Settings/index.jsx
@@ -5,8 +5,10 @@ import get from 'lodash/get'
 
 import { translate } from 'cozy-ui/react/I18n'
 import { Button } from 'cozy-ui/react/Button'
-import { queryConnect } from 'cozy-client/dist'
+import { queryConnect, generateWebLink } from 'cozy-client/dist'
 import { models } from 'cozy-client'
+import flag from 'cozy-flags'
+
 let instanceModel = undefined
 let hasAnOffer = undefined
 let isFremiumFixed = undefined
@@ -92,7 +94,8 @@ export class Settings extends Component {
       contextQuery,
       storageData,
       settingsAppURL,
-      isFetching
+      isFetching,
+      client
     } = this.props
 
     let shouldDisplayViewOfferButton = false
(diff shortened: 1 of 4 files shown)


src/components/Settings/index.jsx
```diff
@@ -114,7 +118,26 @@ export class Settings extends Component {
...
 
-        managerUrlPremiumLink = instanceModel.buildPremiumLink(data)
+        const hasSubscription = flag('settings.subscription')
+
+        if (hasSubscription) {
+          // Contrived but we don't have access to client hooks nor client itself
+          stack.get.capabilities().then(res => {
+            const webLink = generateWebLink({
+              cozyUrl: stack.get.cozyURL(),
+              subDomainType: res.data.attributes.flat_subdomains
```
cballevre, line 128: I think you don't need to get `capabilities` with a request there already into the cozy-client instance you can access them with :
```
const { subdomain: subDomainType } = client.getInstanceOptions()
```
We're trying to do without redux in the cozy-bar as much as possible. To avoid feeding the monster

src/components/Settings/index.jsx
```diff
@@ -114,7 +118,26 @@ export class Settings extends Component {
...
           instanceModel.shouldDisplayOffers(data) || hasAnOffer(data)
 
-        managerUrlPremiumLink = instanceModel.buildPremiumLink(data)
+        const hasSubscription = flag('settings.subscription')
+
+        if (hasSubscription) {
+          // Contrived but we don't have access to client hooks nor client itself
+          stack.get.capabilities().then(res => {
+            const webLink = generateWebLink({
+              cozyUrl: stack.get.cozyURL(),
```
cballevre, line 127: You can access the `cozyUrl` directly with an instance of `CozyClient` with `client.getStackClient().uri`

cballevre (commented):
I admit that the v7 code base isn't ideal, but it would be great if we could avoid making a new query

src/components/Settings/index.jsx
acezard: I'm not sure to understand, I don't think we have access to that in cozy bar v7? only v8

cballevre: I think we can access it because we set a [CozyProvider](https://github.com/cozy/cozy-bar/blob/v7-stable/src/index.jsx#L88) into the `renderBar`. However, I don't know whether there's HOC to access the instance of CozyClient

acezard: I had issues with the HOC because it uses hooks under the hood. So I drilled the prop instead, only one level down so it looks ok to me, what do you think?

cballevre: that seems like a good compromise to me :+1

cballevre (approved):

PR #841, opened 2024-03-19
feat: update managerUrlPremiumLink if sub (VO-386)

acezard:
Will redirect to store if settings.subscription flag is present


src/components/Settings/index.jsx
```diff
@@ -120,8 +124,24 @@ export const Settings = ({
     }
     shouldDisplayViewOfferButton =
       instanceModel.shouldDisplayOffers(data) || hasAnOffer(data)
+    shouldDisplayViewOfferButton =
```
cballevre, line 127: I think this variable duplicates the one above, so it should be removed

cballevre (approved):

PR #858, opened 2024-11-21
fix: apps.hidden flag

paultranvan:


cballevre (approved):
I hadn't seen it either :'(

PR #861, opened 2025-03-13
Move search in bar

zatteo:
Goal : allow every app to enable to search and AI assistant in one line !

See the README commit to understand how it work in this proposition 5a7596f5500eb5daa4fe46ec9b4814108c393b7f

---

To do : 
- [x] update cozy-search

---

In cozy-drive : https://github.com/cozy/cozy-drive/pull/3329
In libs : https://github.com/cozy/cozy-libs/pull/2748

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,36 @@ You also need to disable the internal store onto the `BarComponent` with the pro
 <BarComponent disableInternalStore />
 ```
 
+## Enable search and AI assistant
+
+Search and AI assistant is now proposed by the cozy-bar. To enable it, you need to : 
+
+1. Add the data proxy in your app
+
+- cozy-dataproxy-lib must be installed in your app
+- DataProxyProvider must be added before BarProvider in your app
+
+2. Enable the search
+
+```jsx
+`<BarComponent searchOptions={{ enabled: true }} />`
+```
+
+3. Add the routes
+
+These routes allow to display the AI assistant.
+
+```jsx
+import { BarRoutes } from 'cozy-bar'
+
+<Routes>
+  // Your routes
+  // ...
+
+  {...BarRoutes}
+</Routes>
+```
+
 ## Change theme bar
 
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -60,6 +60,7 @@
     "stylus": "0.57.0"
   },
   "dependencies": {
+    "cozy-search": "file:../cozy-libs/packages/cozy-search",
```
Ldoppea, line 63: Don't forget to set the definitive package reference here

```diff
- "cozy-search": "file:../cozy-libs/packages/cozy-search",
+ "cozy-search": "cozy-search", 
```

README.md
```diff
@@ -89,6 +89,36 @@ You also need to disable the internal store onto the `BarComponent` with the pro
...
 ```
 
+## Enable search and AI assistant
+
+Search and AI assistant is now proposed by the cozy-bar. To enable it, you need to : 
+
+1. Add the data proxy in your app
+
+- cozy-dataproxy-lib must be installed in your app
+- DataProxyProvider must be added before BarProvider in your app
```
Ldoppea, line 99: If we forget this step, do we have a clear error?

Until now, we deployed the search only on a few app and in parallel of the search feature development, so we had control. But now we may want to do the check for better DX.

However I'm not sure if we should throw an error. For now, if no data-proxy, then we just disable the search feature (but AI assistant will work). But the check is in the case the iFrame did not load.
Maybe we can check if the dataProxyProvider is available and throw an error otherwise?

Ldoppea (changes requested):

README.md
zatteo: I added a more explicit message here https://github.com/cozy/cozy-libs/pull/2751

Search disabled and no provider = ok
Search enabled and provider = ok
Search enabled and no provider = blank page but explicit error message with https://github.com/cozy/cozy-libs/pull/2751

Ldoppea (approved):

PR #866, opened 2025-03-31
Adjust new bar style

zatteo:


src/lib/queries.js
```diff
@@ -1,10 +1,13 @@
-import { Q } from 'cozy-client'
+import { Q, fetchPolicies } from 'cozy-client'
+
+const ONE_HOUR = 60 * 60 * 1000
```
JF-Cozy, line 3: if we use realtime on io.cozy.contact, we can rise the fetchpolicies significantly

src/components/AppsMenu/components/AppItem.jsx
```diff
@@ -30,7 +30,12 @@ export const AppItem = ({ app }) => {
             label={
               <div className={styles['apps-menu-grid-item']}>
                 <AppIcon app={app} key={app.slug} {...stack.get.iconProps()} />
-                <Typography variant="caption" noWrap className="u-mt-half">
+                <Typography
+                  color="textPrimary"
+                  variant="caption"
```
JF-Cozy, line 35: i'm not a big fan of using a variant for the size but with a different color, this may be due to the fact that we use more variants than the design system offers... maybe we should talk to Swann about this?

src/styles/apps-menu.styl
```diff
@@ -1,10 +1,14 @@
...
+
 .apps-menu-grid
     display flex
     flex-wrap wrap
     width 100%
     box-sizing border-box
-    padding-left .5rem
-    padding-right .5rem
+
+    +desktop()
```
JF-Cozy, line 9: be careful with this because I think we have to consider the tablet as the desktop 🤔

JF-Cozy (approved):

src/components/AppsMenu/components/AppItem.jsx
zatteo: Agree

src/lib/queries.js
zatteo: It depends of the app using it so I prefer to let it as is.

src/styles/apps-menu.styl
zatteo: It is validated by Swann

PR #867, opened 2025-04-01
Transparent bar for Home app

zatteo:


src/components/Bar.jsx
```diff
@@ -193,7 +195,14 @@ export const Bar = ({
   }
 
   return (
-    <div className="coz-bar-wrapper" data-testid="coz-bar-wrapper">
+    <div
+      {...componentsProps?.CozBarWrapper}
```
JF-Cozy, line 199: we are in CozBar, maybe `componentsProps?.Wrapper` is enough?

JF-Cozy (approved):

src/components/Bar.jsx
zatteo: Deal. Don't like "CozBarWrapper" too.

PR #872, opened 2025-04-11
New sort for app menu & app menu and user menu updates 

zatteo:
**To do** 
- [x] update cozy-client
- [x] update cozy-ui


src/styles/apps-menu.styl
```diff
@@ -33,3 +32,8 @@
     +desktop()
         width 2.625rem
         height 2.625rem
+
+.apps-menu-grid-item-text
```
JF-Cozy, line 36: you should use "makeStyles" (with or without "classes") to override without having to use important, it's the recommande way for Mui component

src/components/AppsMenu/AppsMenuContent.jsx
```diff
@@ -36,13 +35,15 @@ export const AppsMenuContent = ({ isFetchingApps, apps, homeApp }) => {
 
   const homeSlug = homeApp && homeApp.slug
 
-  const displayedApps = apps
-    .filter(app => app.slug !== homeSlug)
-    .sort(sorter(getAppDisplayName))
+  const displayedApps = apps.filter(app => app.slug !== homeSlug)
+
+  const sortedApps = flag('apps.sort')
```
JF-Cozy, line 40: maybe we should export it to use same method in here and Home 🤔

JF-Cozy (approved):

src/components/AppsMenu/AppsMenuContent.jsx
zatteo: `sortApplicationsList` comes from cozy-client applications model. It is enough no ?

PR #878, opened 2025-05-14
feat: Remove isTwakeTheme

zatteo:


JF-Cozy (approved):
I think it shows that we lost ability to custom home icon and rendering somthing in the center of the bar 🤔

PR #881, opened 2025-05-19
fix: Use help link from context settings

zatteo:


src/components/HelpLink/index.jsx
```diff
@@ -17,4 +21,8 @@ const HelpLink = () => {
   )
 }
 
-export default HelpLink
+export const mapStateToProps = state => ({
```
JF-Cozy, line 24: we can use hook instead:

```
import { useSelector } from 'react-redux'
const helpLink = useSelector(getHelpLink)`
```

JF-Cozy (approved):

src/components/HelpLink/index.jsx
zatteo: Much better 👍

PR #882, opened 2025-06-11
Remove old bar

zatteo:


package.json
JF-Cozy, line 1: is this whole commit a BC? At least I think it's a feat instead of refactor

zatteo: Agree I will mark it as a BC

src/components/utils/ButtonCozyHome.spec.jsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react'
 import { render, fireEvent } from '@testing-library/react'
-import { ButtonCozyHome } from './ButtonCozyHome'
+import ButtonCozyHome from './ButtonCozyHome'
```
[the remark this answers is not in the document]
zatteo: Keeping it like this for the moment, not a priority.

PR #885, opened 2025-10-07
feat: Display entrypoints in cozy-bar

zatteo:


src/components/AppsMenu/helper.js
```diff
@@ -0,0 +1,10 @@
+export const buildEntrypoints = apps => {
+  return apps.flatMap(app => {
+    if (!app.entrypoints) return []
+
+    return app.entrypoints.map(entrypoint => ({
+      ...entrypoint,
+      slug: app.slug
+    }))
+  })
+}
```
zatteo, line 10: I hesitated to centralize this code on cozy-client, but as entrypoints are not really mature (specification already changed between yesterday and today), I think it is too soon for this kind of centralization.

PR #902, opened 2026-03-23
Remove redux from cozy-bar

zatteo:
- Removed lot of outdated code
- Removed lot of outdated dependencies
- Same feature
- Less complexity
- -68KB for build twake-drive build size (uncompressed)

Tested on Drive (private and public page) and a container app.


src/components/helpers.js
```diff
@@ -0,0 +1,27 @@
...
+
+// cozy-client's useQuery normalizes JSON:API responses — attributes are
+// spread to the top level (e.g. app.slug, app.name exist directly).
+// We only ensure `href` is derived from `links.related` for AppLinker.
+const mapApp = app => ({
+  ...app,
+  href: (app.links && app.links.related) || app.href
+})
+
+export const getAppsData = (rawApps, appSlug) => {
```
JF-Cozy, line 11: nit: sounds complicated 🤔 is it because we don't use apps data correctly in our components ? Maybe some refactoring can help to cleanup a bit 🤷‍♂️

JF-Cozy (approved):

src/components/helpers.js
zatteo: We use:
- a filtered version of apps
- isSettingsAppInstalled
- homeApp => this one could be removed for simplicity. I will have other work to do on cozy-bar so I will see.

